PR #27, opened 2019-02-12
PR to make USD a package

clusty:
Most are file moves.
Only code changes are:
* InitUsd.cs to init plugins from the new path
* Changed UsdAsset.m_FilePath to be private since to load assets that are in packages, you need to call **System.IO.Path.GetFullPath** to get the real path on disk.


package/com.unity.formats.usd/Runtime/Scripts/IO/Skel/SkeletonImporter.cs
```diff
@@ -15,6 +15,7 @@
 using System;
 using System.Collections.Generic;
 using pxr;
+using Unity.Collections;
```
jcowles: Unity.Collections does not exist in older versions of Unity -- hence the conditional include a few lines below this.

package/com.unity.formats.usd/Runtime/Scripts/Behaviors/UsdAsset.cs
```diff
@@ -30,9 +31,15 @@ public class UsdAsset : MonoBehaviour {
     public double Length { get { return ComputeLength(); } }
 
     [Header("Source Asset")]
-    public string m_usdFile;
-    public string m_projectAssetPath = "Assets/";
+    [SerializeField]
+    string m_usdFile;
 
+    public string fullPath
```
jcowles: I feel like "fullPath" is kind of ambiguous -- I think we should either add a doc comment explaining that it's the full path to the USD file or rename to usdFileFullPath? What do you think?

package/com.unity.formats.usd/Runtime/Scripts/IO/Skel/SkeletonImporter.cs
clusty: Done

package/com.unity.formats.usd/Runtime/Scripts/Behaviors/UsdAsset.cs
clusty: renamed to usdFullPath

package/com.unity.formats.usd/package.json
```diff
@@ -0,0 +1,19 @@
+{
+    "dependencies": {
+        "com.unity.timeline" : "1.0.0"
+    },
+    "description": "The USD package provides an API to manipulate Universal Scene Description (.usd) files. USD is a 3D Scene format that scales well to large scenes and contains assets of different types (geometry, images, etc).",
```
mfe: For approved description / tags : 
see : https://docs.google.com/document/d/1jhRFrI3PRp4GvgpVjl2Sa7CxC06jFhX2WuPKwfA-HV4/edit#

package/com.unity.formats.usd/Runtime/Unity.Formats.USD.Runtime.asmdef
```diff
@@ -0,0 +1,18 @@
+{
+    "name": "Unity.Formats.USD.Runtime",
+    "references": [
+        "Unity.Timeline"
+    ],
+    "optionalUnityReferences": [],
+    "includePlatforms": [
+        "Editor",
+        "macOSStandalone",
```
mfe: Logged an issue for the failing build : https://unity3d.atlassian.net/browse/FTV-75

package/com.unity.formats.usd/package.json
```diff
@@ -0,0 +1,19 @@
...
+    ], 
+    "name": "com.unity.formats.usd",
+    "repoPackagePath": "package/com.unity.formats.usd", 
+    "repository": {
+        "type": "git", 
+        "url": "ssh://git@github.com:Unity-Technologies/usd-unity-sdk.git"
+    }, 
+    "unity": "2018.3",
+    "unityRelease": "4f1",
+    "version": "0.0.1-preview.1"
```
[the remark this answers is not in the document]
jcowles: No objections.
Note that the .NET DLLs are versioned as 19.01, which matches USD. Making these match the package version probably makes most sense, however it will cause version numbers to go down, which will likely throw some warnings for existing users.

PR #32, opened 2019-02-19
Convert Example into package Samples

clusty:
Convert Examples into Package samples. This is available since 2019.1
Fix null/empty strings in UsdAsset


mfe:
It is super nice to be able to import those samples in 2019.1 :)
But how can a 2018.3 user access them ? As they are not visible in the package and Packman UI does not yet handle the imports in this version.

clusty:
for under 2019.1, we'll have to figure out another way for them to get their Samples: @jcowles suggested GitHub, you suggested maybe the film samples.

mfe:
So after discussing with Vlad : 
* Put samples on [film-tv-toolbox](https://github.com/Unity-Technologies/film-tv-toolbox) for < 2019.1 users,
* Update README to add document where the samples are depending on Unity version.

package/com.unity.formats.usd/Samples~/ImportMaterials/ImportMaterialsExample.cs
```diff
@@ -241,7 +241,7 @@ public struct ShaderPair {
       var detailNormalMapPath = "/Model/Materials/SimpleMat/DetailNormalMap";
       var detailMaskPath = "/Model/Materials/SimpleMat/DetailMask";
 
-      var textureFilePath = @"Assets/UsdUnitySdk/Examples/ImportMaterials/Textures/";
+      var textureFilePath = @"Assets/Samples/USD/1.0.0-preview.1/ImportMaterials/Textures/";
```
mfe: For later, find a way to not have the version baked in the code.

PR #33, opened 2019-02-20
Initial Test and Continuous Integration with Yamato

clusty:
*Converted HelloUsd into a test (the most trivial USD operation I found around)
*Added a dummy Documentation~/USD.md with same info from package.json (needed this to make validation green)
*Added Yamato CI config. You can run it at: https://yamato.prd.cds.internal.unity3d.com/jobs/86-usd-unity-sdk . 2019.1 will fail on Mac because of a bug fix in Unity that did not make it in just yet.

### DIFF
--- a/.yamato/upm-ci.yml
+++ b/.yamato/upm-ci.yml
@@ -1,6 +1,7 @@
 editors:
   - version: 2019.1.0b2
-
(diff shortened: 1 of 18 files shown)


jcowles:
Nice!

PR #38, opened 2019-03-01
Support 'Compute' for MeshImporter.color

TimsLuxo:
MeshImporter.color=Compute forces constant to vertex and uses first color as vertex color if other methods fail


jcowles (approved):

jcowles:
Holding this to merge until after preview1 has been released

TimsLuxo:
> Holding this to merge until after preview1 has been released

If possible, please merge this, as we have several use cases that currently require fragile workarounds. Thanks.

jcowles:
Doh, sorry to leave this hanging so long!

jcowles:
Hey @TimsLuxo, this new compute code path only gets hit when the mesh colors are broken or are uniform, right?

Are you actually just trying to add support for uniform colors?

package/com.unity.formats.usd/Runtime/Scripts/IO/Geometry/MeshImporter.cs
```diff
@@ -506,13 +512,29 @@ public class GeometrySubsets {
...
                 new Exception("Error unrolling Face-Varying colors at <" + path + ">", ex));
           }
         } else {
           Debug.LogWarning("Uniform (color per face) display color not supported");
         }
+
+        if (forceVertexColors) {
+            Color [] colors = new Color[usdMesh.points.Length];
+            if (usdMesh.colors.Length != usdMesh.points.Length) {
+              Color color = usdMesh.colors[0].gamma;
```
jcowles: Do you actually want `gamma` here?

I think we only want the color in gamma space when it's used as a material parameter, and now I'm questioning if that's even correct for HDRP/UniversalRP...

TimsLuxo: Good point. I suppose not in this case. Likewise with *RP which are linear, correct?

jcowles: Yeah, I just confirmed the SRP pipelines are linear, so we have a bug in the other code calling .gamma above.

jcowles (changes requested):
Would be nice to kill the .gamma call (ultimate fix TBD) and perhaps refactor this to just be the default when we find uniform colors?

TimsLuxo:
> Would be nice to kill the .gamma call (ultimate fix TBD) and perhaps refactor this to just be the default when we find uniform colors?

This is implemented in latest update to pull request.

jcowles:
I'm sincerely sorry for sitting on this for so long @TimsLuxo - I'm going to close this PR, since #261 refactors so much of the import code.

If you ever want to revisit this, ping me and I'll make it happen.

PR #45, opened 2019-03-12
Root prim fix

jcowles:
The old logic would always attempt to set a root prim, the new logic only sets the root when only a single root exists. This also moves the variability cache option into the debug section of the UsdAsset inspector.

### DIFF
--- a/package/com.unity.formats.usd/Runtime/Scripts/Timeline/UsdPlayableAsset.cs
+++ b/package/com.unity.formats.usd/Runtime/Scripts/Timeline/UsdPlayableAsset.cs
@@ -24,12 +24,11 @@ namespace Unity.Formats.USD {
 
     [Tooltip("USD Player to Control")]
     public ExposedReference<UsdAsset> SourceUsdAsset;
-    public string UsdRootPath;
 
     public ClipCaps clipCaps { get { return ClipCaps.Extrapolation | ClipCaps.Looping | ClipCaps.SpeedMultiplier | ClipCaps.ClipIn; } }
 
-    public UsdAsset GetUsdAsset() {
-      m_sourceUsdAsset.m_usdRootPath = UsdRootPath;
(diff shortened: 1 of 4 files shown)


package/com.unity.formats.usd/Runtime/Scripts/Timeline/UsdPlayableAsset.cs
```diff
@@ -24,12 +24,11 @@ public class UsdPlayableAsset : PlayableAsset, ITimelineClipAsset {
...
     [Tooltip("USD Player to Control")]
     public ExposedReference<UsdAsset> SourceUsdAsset;
-    public string UsdRootPath;
 
     public ClipCaps clipCaps { get { return ClipCaps.Extrapolation | ClipCaps.Looping | ClipCaps.SpeedMultiplier | ClipCaps.ClipIn; } }
 
-    public UsdAsset GetUsdAsset() {
-      m_sourceUsdAsset.m_usdRootPath = UsdRootPath;
+    public UsdAsset GetUsdAsset(string usdRootPath) {
+      m_sourceUsdAsset.m_usdRootPath = usdRootPath;
```
clusty: isn't this a strange behaviour ? Normally I'd expect getters to be "const". Does USD rely on this behaviour? If not we could return a copy of the m_sourceUsdAsset with the  m_usdRootPath set at the correct value.

jcowles: There is no change in behavior, but yes, I agree, it sucks.
This was not by choice, it was because the Timeline API doesn't seem to expose the target of the track at the point when clips are created.

If it is possible to this (e.g. on line 43, below), I'm all for it. Otherwise, I don't see how it can be avoided, aside from perhaps only setting the root path once.

PR #65, opened 2019-03-16
SetHideFlags now ORs HideFlags rather than setting. Changed defaults …

TimsLuxo:
…work better with default Usd Asset import settings. The old code overrode the Hide settings for instance Masters, making them visible in the Hierarchy scene graph.

### DIFF
--- a/package/com.unity.formats.usd/Samples/ImportProcessor/SetHideFlags.cs
+++ b/package/com.unity.formats.usd/Samples/ImportProcessor/SetHideFlags.cs
@@ -25,7 +25,7 @@ namespace Unity.Formats.USD.Examples {
     public HideFlags hideFlagsSettings = HideFlags.DontSave;
 
     public enum Operation {Overwrite, LogicalAND, LogicalOR, LogicalXOR}
-    [Tooltip("The logical operator to use when setting the flags. In most cases Or will produce expectd results")]
+    [Tooltip("The logical operator to use when setting the flags. In most cases Or will produce expected results")]
     public Operation operation = Operation.LogicalOR;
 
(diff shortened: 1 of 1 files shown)


package/com.unity.formats.usd/Samples/ImportProcessor/SetHideFlags.cs
```diff
@@ -24,21 +24,38 @@ public class SetHideFlags : RegexImportProcessor, IImportPostProcessHierarchy {
 
     public HideFlags hideFlagsSettings = HideFlags.DontSave;
 
+    public enum Operation {Overwrite, LogicalAND, LogicalOR, LogicalXOR}
+    [Tooltip("The logical operator to use when setting the flags. In most cases Or will produce expectd results")]
```
jcowles: Typo in "expectd"

package/com.unity.formats.usd/Samples/ImportProcessor/SetHideFlags.cs
```diff
@@ -24,21 +24,38 @@ public class SetHideFlags : RegexImportProcessor, IImportPostProcessHierarchy {
...
+          default:
+            go.hideFlags = hideFlagsSettings;
+            return;
+        }
       }
     }
 
     void Reset() {
-      matchExpression = "/";
+      matchExpression = "/World";
```
jcowles: Why "/World"?

TimsLuxo:
If it’s set to / then donotsave is applied to the root, meaning the root
will disappear on save/reload.
Probably should be changed to use mode regex and “/.+”

jcowles:
Sorry for the delay, LGTM!

PR #79, opened 2019-04-09
USD: Upgrade to 19.05

jcowles:
Currently this is built using the USD release candidate, I will update the PR when the final release is cut.

This also does not yet include OSX libs, which will cause the mac build to fail due to changes in Pcp.


jcowles:
Currently this is built using the USD release candidate, I will update the PR when the final release is cut.

This also does not yet include OSX libs, which will cause the mac build to fail due to changes in Pcp.

PR #83, opened 2019-04-23
Linux cmake

mfe:


CMakeLists.txt
```diff
@@ -0,0 +1,201 @@
+cmake_minimum_required(VERSION 3.8)
+
+project(usd-unity-sdk LANGUAGES CXX)
+set (CMAKE_CXX_STANDARD 11)
+list(APPEND CMAKE_MODULE_PATH ${CMAKE_SOURCE_DIR}/cmake)
+
+find_package(pxr REQUIRED)
+find_package(SWIG 3.0.12 REQUIRED)
+find_package(Boost REQUIRED)
```
clusty: @cpredmann Does this find the package installed on the OS ? If yes, I think we should control much more tightly the dependency versions to be able to be more reproducible (download any dependencies as a part of the built process)

CMakeLists.txt
```diff
@@ -0,0 +1,201 @@
...
+find_package(Python COMPONENTS Development)
+find_package(TBB REQUIRED)
+find_package(Unity REQUIRED)
```
clusty: @cpredmann do you need both Unity and a C# compiler ?

cpredmann: The FindUnity.cmake module also gets the C# compiler (in addition to Unity), and it does need both, since the USD.NET.Unity .dll references UnityEngine.dll.

CMakeLists.txt
cpredmann: Yes, it'll just look for those modules, but you're right that it doesn't (aside from SWIG) do any version checking.  I could add ExternalProject_Add steps for the dependencies to make sure it's all built correctly (version-wise).  That would also solve the problem of potentially dynamically linking against random versions of dependencies, if they were built/installed as part of the process.

CMakeLists.txt
clusty: Might it make sense to add the USD.NET.Unity dll as C# files in the package and let Unity compile them every time ?

cpredmann: I'd lean towards just compiling the dll at build, since the overwhelming majority are just Swig generated and don't make much sense out of that context

cpredmann: Although that's for the USD.NET.dll, so through C# for the .Unity dll could be packaged as is

clusty: basically what I am getting at is that requiring a unity installation makes for a very CI unfriendly setup, if ever you want to automate the compilation of USD and/or dependencies.

cpredmann: Got it... so in that case compiling the Swig generated C# only requires a mono compiler, so it may make sense to build the USD.NET.dll, but not the USD.NET.Unity.dll (which is the only thing requiring a Unity install)

cpredmann:
I've added in build steps for the dependencies, and removed the need for Unity.  Testing with moving the USD.NET.Unity C# files into the package without compilation, I needed to add unsafe to the asmdef, but works otherwise (on Linux).

CMakeLists.txt
```diff
@@ -4,43 +4,109 @@ project(usd-unity-sdk LANGUAGES CXX)
...
+   Alembic
+   GIT_REPOSITORY https://github.com/alembic/alembic.git
+   GIT_TAG 1.7.7
+   GIT_SHALLOW TRUE
+   CMAKE_ARGS -DCMAKE_INSTALL_PREFIX=${dep_path}/Alembic -DUSE_BINARIES=OFF -DUSE_TESTS=OFF -DCMAKE_BUILD_TYPE=Release
+)
+
+set(USD_EXE_LINKER_FLAGS -L${dep_path}/Alembic/lib)
+list(APPEND USD_EXE_LINKER_FLAGS -lAlembic)
+ExternalProject_Add(
```
clusty: Really cool stuff. Did not know you could do that

clusty:
I pulled from the fork and pushed it on the main repo to see yamato. I think this is missing some build script updates and we need to see if the yamato images have all the required dependencies.

jcowles:
Adding USD.NET to the package will make package compilation take a lot longer.

Also, there are users of USD.NET outside of Unity.

What's he motivation for including it in the package?

CMakeLists.txt
jcowles: +1, versions are specific and should stay in lock step with the USD build script dependency versions (which themselves are driven by the VFX Platform).

cpredmann: Yes, the updated version of the CMake builds with the specific version dependencies, to avoid any compilation issues/conflicts.

PR #85, opened 2019-04-26
FTV-143

cpredmann:
* Add all platforms, static linking

* Add USD.NET.Unity C#


jcowles:
It looks like the USD.NET project is now in the package, was that intended?

cpredmann:
It should just be USD.NET.Unity, but not USD.NET.  This was in response to @clusty's comment about excluding Unity from the build requirements...all on this branch is very much in progress, and I'm certainly looking for feedback and input.

clusty:
it was in the optics that the CI would build everything every time (since binaries were remove). having a dependency on unity would make yamato build very hard. I think we decided to stick with committing precanned binaries for now (so my comment is obsolete for now)

PR #87, opened 2019-05-06
Linux support (FTV-143)

cpredmann:
This adds Linux support for the package, along with a CMake build configuration for the Swig library (only Linux supported currently).  The CMake is optionally able to build Mono and the USD.NET.dll, but this is disabled by default.  None of the Windows or OSX configurations have changed, and this was tested on 2018.3 and 2019.1.


jcowles:
Hrm, if this is based on dev, why did the Yamato config fix need to be pulled in?

cpredmann:
This had been based off of master, which is why it was missing that (along with 19.05 updates).  I've rebased against dev, and updated the Linux binary and CMake to use 19.05

cmake/CMakeLists.txt
```diff
@@ -0,0 +1,286 @@
+cmake_minimum_required(VERSION 3.13)
+
+project(usd-unity-sdk LANGUAGES CXX)
+set(CMAKE_CXX_STANDARD 11)
+
+set(USD_VERSION "v19.05" CACHE STRING "USD version (git tag)")
+
+if (${CMAKE_SYSTEM_NAME} STREQUAL "Linux")
+    set(boost_url "https://downloads.sourceforge.net/project/boost/boost/1.55.0/boost_1_55_0.tar.gz")
```
clusty: 2 different boost versions for: Windows/Unix

clusty:
Is the current Howto up to date on how to update the USD version ?

cmake/CMakeLists.txt
cpredmann: That's just following along with the versions from the USD build script (https://github.com/PixarAnimationStudios/USD/blob/38c70fb869c01228f82fd844a8c420860b5131c1/build_scripts/build_usd.py#L519), but it would be nice to figure out why they're not moving everything to VFX CY2018 (at least).

jcowles:
I don't know CMake well, but I'm assuming Vlad does since he commented on it.

And yeah, if you wouldn't mind updating the README.md and BUILDING.MD, I think that's all that's left.

Thanks Chris!!

cpredmann:
Sure thing!  I made a couple of edits to the README and BUILDING md files, if you want to take a look over...

jcowles:
(I just rebased this branch to dev, FYI)

PR #95, opened 2019-06-11
Playback: Fix data race

### LINKED ISSUE #93: Timeline: Playback sometimes spews errors (FTV-195)
It seems there is an off-by-one error somewhere in the playback job system, which results in very infrequent error messages about null objects.

I have a fix.

jcowles:
Previously, streaming playback was relying on reading and writing bool[] elements to be atomic, however bool arrays are implemented as byte arrays, which means reading and writing is not atomic and the elements are aliased.

This fixes the issue by checking the reference array and removing the "written" bool array entirely.

Fixes #93


jcowles:
I think I'd like to make one more change here -- it's still using a List<bool> in a multi-threaded context, which I want to take a closer look at.

PR #96, opened 2019-06-11
Material Import: Triangulate mesh subsets before binding materials

jcowles:
Previously the geometry subset face indices were not triangulated, which meant the indices were references to n-gons.

With this change, the indices are triangulated, which fixes previously broken material bindings.

### DIFF
--- a/package/com.unity.formats.usd/Runtime/Scripts/IO/Geometry/MeshImporter.cs
+++ b/package/com.unity.formats.usd/Runtime/Scripts/IO/Geometry/MeshImporter.cs
@@ -351,7 +351,7 @@ namespace Unity.Formats.USD {
 
           // UsdGeomSubsets contain a list of face indices, but once the mesh is triangulated, these
           // indices are no longer correct. The number of new faces generated is a fixed function of
-          // the original face indices, but the offset requres an accumulation of all triangulated
+          // the original face indices, but the offset requires an accumulation of all triangulated
           // face offsets, this is "offsetMapping". The index into offsetMapping is the original
           // face index and the value at that index is the new face offset.
           //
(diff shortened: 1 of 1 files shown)


package/com.unity.formats.usd/Runtime/Scripts/IO/Geometry/MeshImporter.cs
```diff
@@ -348,6 +348,33 @@ public class GeometrySubsets {
...
+          // TODO: this should be moved to C++
+          int[] offsetMapping = new int[usdMesh.faceVertexCounts.Length];
+          int curOffset = 0;
+          for (int i = 0; i < usdMesh.faceVertexCounts.Length; i++) {
+            offsetMapping[i] = curOffset;
+            curOffset += Math.Max(0, usdMesh.faceVertexCounts[i] - 3);
+          }
+
+          var newSubsets = new GeometrySubsets();
+          foreach (var kvp in geomSubsets.Subsets) {
```
lucillecaillaud: What does "kvp" means (key, value, something?)? Would it be clearer to call the variable "subset"?

package/com.unity.formats.usd/Runtime/Scripts/IO/Geometry/MeshImporter.cs
```diff
@@ -348,6 +348,33 @@ public class GeometrySubsets {
           var counts = UnityTypeConverter.ToVtArray(usdMesh.faceVertexCounts);
           UsdGeomMesh.Triangulate(indices, counts);
           UnityTypeConverter.FromVtArray(indices, ref usdMesh.faceVertexIndices);
+
+          // UsdGeomSubsets contain a list of face indices, but once the mesh is triangulated, these
+          // indices are no longer correct. The number of new faces generated is a fixed function of
+          // the original face indices, but the offset requres an accumulation of all triangulated
```
lucillecaillaud: You've got a small typo here on "requres"

package/com.unity.formats.usd/Runtime/Scripts/IO/Geometry/MeshImporter.cs
jcowles: "key-value pair" I could call it nameAndSubset

package/com.unity.formats.usd/Runtime/Scripts/IO/Geometry/MeshImporter.cs
jcowles: Will fix!

PR #99, opened 2019-06-21
Fix the ExportUsdz function to get every archive inner paths relative (FTV-172, #88)

lucillecaillaud:
Use the native USD function: UsdUtilsCreateNewARKitUsdzPackage. This function find all the dependencies for us and archive everything together.

To ensure that all path were relative, the archive had to be created in the same folder were the source usd file and it's dependencies was. The archive is then copied to the destination folder provided.

### DIFF
--- a/package/com.unity.formats.usd/Runtime/Scripts/IO/Scene/UsdzExporter.cs
+++ b/package/com.unity.formats.usd/Runtime/Scripts/IO/Scene/UsdzExporter.cs
@@ -15,45 +15,35 @@
 using System.IO;
 using UnityEngine;
 using USD.NET;
+using pxr;
 
 namespace Unity.Formats.USD {
   public class UsdzExporter {
 
     public static void ExportUsdz(string usdzFilePath,
                                   GameObject root) {
-      // Setup a temp directory for zipping up files.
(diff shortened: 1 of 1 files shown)


package/com.unity.formats.usd/Runtime/Scripts/IO/Scene/UsdzExporter.cs
```diff
@@ -72,38 +62,26 @@ public class UsdzExporter {
...
-          if (!supportedExtensions.Contains(ext)) {
-            Debug.LogWarning("Unsupported file type in USDZ: " + relPath);
-            continue;
-          }
-          filesToArchive.Add(relPath);
+        SdfAssetPath assetPath = new SdfAssetPath(usdcFileName);
+        bool success = pxr.UsdCs.UsdUtilsCreateNewARKitUsdzPackage(assetPath, usdzFileName);
+
+        if (!success) {
+          Debug.LogError("Couldn't export " + root.name + " to the usdz file: " + usdzFilePath);
```
jcowles: Should probably exit here / not try to copy the file after failure

jcowles:
LGTM!

PR #103, opened 2019-07-15
VariantSet at root fix

### LINKED ISSUE #77: VariantSets on the root UsdAsset results in errors (FTV-203)
The issue is the UsdAsset itself has no UsdPrimSource (though it should), so when the variant selection state updates, it accesses the UsdPrimSource and causes a null reference exception.

jcowles:
Fixes #77

### DIFF
--- a/package/com.unity.formats.usd/Runtime/Scripts/Behaviors/UsdAsset.cs
+++ b/package/com.unity.formats.usd/Runtime/Scripts/Behaviors/UsdAsset.cs
@@ -489,7 +489,7 @@ namespace Unity.Formats.USD {
     /// </summary>
     public void DestroyAllImportedObjects() {
       foreach (var src in GetComponentsInChildren<UsdPrimSource>(includeInactive: true)) {
-        if (src) {
+        if (src && src.gameObject != this.gameObject) {
           GameObject.DestroyImmediate(src.gameObject);
         }
       }
@@ -787,6 +787,11 @@ namespace Unity.Formats.USD {
     /// be destroyed and imported as a result.
     /// </summary>
     private void ApplyVariantSelectionState(Scene scene, UsdVariantSet variants) {
+      var primSource = variants.GetComponent<UsdPrimSource>();
+      if (!primSource) {
+        Debug.LogError("Null UsdPrimSource while applying variant selection state.");
+        return;
+      }
       var selections = variants.GetVariantSelections();
       var path = variants.GetComponent<UsdPrimSource>().m_usdPrimPath;
       var prim = scene.GetPrimAtPath(path);
--- a/package/com.unity.formats.usd/Runtime/Scripts/Behaviors/UsdAsset.cs
+++ b/package/com.unity.formats.usd/Runtime/Scripts/Behaviors/UsdAsset.cs
@@ -489,6 +489,9 @@ namespace Unity.Formats.USD {
     /// </summary>
     public void DestroyAllImportedObjects() {
       foreach (var src in GetComponentsInChildren<UsdPrimSource>(includeInactive: true)) {
+        // Remove the object if it is valid, but never remove the UsdAsset root GameObject, which
+        // is this.gameObject. The current workflow is to remove all objects EXCEPT the root, so
+        // stubs of USD Assets can be left in the scene in the scene and imported only as needed.
         if (src && src.gameObject != this.gameObject) {
           GameObject.DestroyImmediate(src.gameObject);
         }
@@ -793,7 +796,7 @@ namespace Unity.Formats.USD {
         return;
       }
       var selections = variants.GetVariantSelections();
-      var path = variants.GetComponent<UsdPrimSource>().m_usdPrimPath;
+      var path = primSource.m_usdPrimPath;
       var prim = scene.GetPrimAtPath(path);
       var varSets = prim.GetVariantSets();
       foreach (var sel in selections) {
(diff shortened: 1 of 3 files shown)


package/com.unity.formats.usd/Runtime/Scripts/Behaviors/UsdAsset.cs
```diff
@@ -787,6 +787,11 @@ public string usdFullPath
     /// be destroyed and imported as a result.
     /// </summary>
     private void ApplyVariantSelectionState(Scene scene, UsdVariantSet variants) {
+      var primSource = variants.GetComponent<UsdPrimSource>();
+      if (!primSource) {
+        Debug.LogError("Null UsdPrimSource while applying variant selection state.");
+        return;
+      }
       var selections = variants.GetVariantSelections();
       var path = variants.GetComponent<UsdPrimSource>().m_usdPrimPath;
```
lucillecaillaud: You already have `primSource` above, I don't think you need to re-do `variants.GetComponent<UsdPrimSource>()`

package/com.unity.formats.usd/Runtime/Scripts/Behaviors/UsdAsset.cs
```diff
@@ -787,6 +787,11 @@ public string usdFullPath
     /// be destroyed and imported as a result.
     /// </summary>
     private void ApplyVariantSelectionState(Scene scene, UsdVariantSet variants) {
+      var primSource = variants.GetComponent<UsdPrimSource>();
```
lucillecaillaud: Question, the `GetComponent` method isn't something on `GameObject`? But `variants` is a `UsdVariantSet`... Is there something that makes USD object be somewhat compatible with Unity object with a simili shared interface? Some magic somewhere ^^

package/com.unity.formats.usd/Runtime/Scripts/Behaviors/UsdAsset.cs
```diff
@@ -489,7 +489,7 @@ public string usdFullPath
     /// </summary>
     public void DestroyAllImportedObjects() {
       foreach (var src in GetComponentsInChildren<UsdPrimSource>(includeInactive: true)) {
-        if (src) {
+        if (src && src.gameObject != this.gameObject) {
```
lucillecaillaud: Could you precise why you added this check? `this.gameObject` isn't part of "AllImportedObjects"?

package/com.unity.formats.usd/Runtime/Scripts/IO/Scene/SceneImporter.cs
```diff
@@ -309,6 +319,16 @@ public ImportException(string message, System.Exception innerException)
...
       if (importOptions.importHierarchy || importOptions.forceRebuild) {
+        // TODO(jcowles): This feels like a workaround. What we really want here is an "undo"
+        // process for changes made to the root GameObject. For example, to clean up non-USD
+        // components which may have been added.
+        RemoveComponent<UsdAssemblyRoot>(root);
+        RemoveComponent<UsdVariantSet>(root);
+        RemoveComponent<UsdModelRoot>(root);
+        RemoveComponent<UsdLayerStack>(root);
+        RemoveComponent<UsdPayload>(root);
+        RemoveComponent<UsdPrimSource>(root);
```
lucillecaillaud: I'm not sure why you added all this `RemoveComponent` thing and how it is linked to the original problem?

package/com.unity.formats.usd/Runtime/Scripts/Behaviors/UsdAsset.cs
jcowles: Good catch - fixed

package/com.unity.formats.usd/Runtime/Scripts/Behaviors/UsdAsset.cs
jcowles: This is a confusing naming convention -- UsdVariantSet in this case is a MonoBehaviour, so it too has a GetComponent method.

We should consider renaming everything in the Runtime/Scripts/Behaviors folder, it's currently very confusing :/

package/com.unity.formats.usd/Runtime/Scripts/Behaviors/UsdAsset.cs
jcowles: Good call -- done

jcowles: I added a comment in the code, but to explain here: the current game object is the UsdAsset root object. If it is removed, then the entire imported object will be gone. The current thought is to remove all objects EXCEPT the root, so that you can have a workflow where you leave stubs of USD Assets in the scene and import them only as needed.

package/com.unity.formats.usd/Runtime/Scripts/IO/Scene/SceneImporter.cs
jcowles: When a USD file is fully RE-imported, all exsiting USD data must be removed. The old assumption was that the root would never have much more than the UsdAsset component itself, however it's now clear that the root may also have meaningful USD data added too.

This is an attempt to clean the root UsdAsset GameObject on re-import, however as I mentioned in the TODO comment, it it is incomplete (imagine if a mesh is imported to the root) and we really need a full "undo" operation to clean the root GameObject. Unfortunately, this is very difficult currently.

I'll add this comment in the code as well!

PR #108, opened 2019-07-16
Samples: add proper asmdef to avoid build errors (FTV-191)

mfe:
When importing some samples, build could fail because of missing definitions.
Fix consisted into adding asmdef files that reference the precompiled libs (USD.Net).

### DIFF
--- a/package/com.unity.formats.usd/Samples/ImportMaterials/Unity.Formats.USD.ImportMaterials.asmdef
+++ b/package/com.unity.formats.usd/Samples/ImportMaterials/Unity.Formats.USD.ImportMaterials.asmdef
@@ -1,0 +1,24 @@
+{
+    "name": "Unity.Formats.USD.ImportMaterials",
+    "references": [
+        "Unity.Formats.USD.Runtime"
+    ],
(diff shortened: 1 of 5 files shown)


package/com.unity.formats.usd/Samples/ImportMaterials/Unity.Formats.USD.ImportMaterials.asmdef
```diff
@@ -0,0 +1,24 @@
...
+        "LinuxStandalone64",
+        "macOSStandalone",
+        "WindowsStandalone64"
+    ],
+    "excludePlatforms": [],
+    "allowUnsafeCode": false,
+    "overrideReferences": true,
+    "precompiledReferences": [
+        "USD.NET.dll",
+        "USD.NET.Unity.dll"
```
jcowles: Is it the case that only this one sample needs USD.NET.Unity.dll?

mfe: Yes

PR #119, opened 2019-07-26
Include mesh tangents in basis conversion on export

samuelschmidt-brinx:
I noticed my normal maps were rendering incorrectly when doing a round trip from Unity > USD > Unity. I eventually narrowed it down to MeshExporter.cs not converting the basis for mesh tangents when using slow and safe transformation. I copied the basis transformation process used to convert the tangents on import into the export step where the vertices and normals are transformed. Now, tangents seem to export correctly and normal maps seem to render correctly. Attached is a screenshot demonstrating the difference with a shader that draws the mesh's tangents as color - left is the original model, middle is the model round-tripped with the fix and right is the model round-tripped without the fix.
Unity_mX8M1tTrvR

### DIFF
--- a/package/com.unity.formats.usd/Runtime/Scripts/IO/Geometry/MeshExporter.cs
+++ b/package/com.unity.formats.usd/Runtime/Scripts/IO/Geometry/MeshExporter.cs
@@ -239,6 +239,11 @@ namespace Unity.Formats.USD {
             if (sample.normals != null && sample.normals.Length == sample.points.Length) {
               sample.normals[i] = UnityTypeConverter.ChangeBasis(sample.normals[i]);
             }
+			if (sample.tangents != null && sample.tangents.Length == sample.points.Length) {
+              var w = sample.tangents[i].w;
+              var t = UnityTypeConverter.ChangeBasis(sample.tangents[i]);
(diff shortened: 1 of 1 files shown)


jcowles:
Nice catch and thank you for the PR!

package/com.unity.formats.usd/Runtime/Scripts/IO/Geometry/MeshExporter.cs
```diff
@@ -238,6 +238,11 @@ public static class MeshExporter {
             sample.points[i] = UnityTypeConverter.ChangeBasis(sample.points[i]);
             if (sample.normals != null && sample.normals.Length == sample.points.Length) {
               sample.normals[i] = UnityTypeConverter.ChangeBasis(sample.normals[i]);
+            }
+			if (sample.tangents != null && sample.tangents.Length == sample.points.Length) {
```
jcowles: Looks like a tabs vs. spaces issue here.

samuelschmidt-brinx: Ah, that's what I get for switching text editors for a moment. I have pushed a fix.

PR #123, opened 2019-08-10
MeshExporter mesh readable fix in editor

soraryu:
Meshes in editor are always readable, no need to return. Also, changed Log from warning to error as this actually corrupts data when it fails

### DIFF
--- a/package/com.unity.formats.usd/Runtime/Scripts/IO/Geometry/MeshExporter.cs
+++ b/package/com.unity.formats.usd/Runtime/Scripts/IO/Geometry/MeshExporter.cs
@@ -154,24 +154,24 @@ namespace Unity.Formats.USD {
         // This is a workaround for a Unity peculiarity - 
         // non-readable meshes are actually always accessible from the Editor.
         // We're still logging a warning since this won't work in a build.
-        static bool CanReadMesh(Mesh mesh)
-        {
-            if (mesh.isReadable) return true;
-            if (!Application.isPlaying) return true;
-
-            if (Mesh_canAccess == null)
+        static bool CanReadMesh(Mesh mesh) {
+            if (mesh.isReadable) {
+                return true;
+            }
+            if (!Application.isPlaying) {
+                return true;
+            }
+            if (Mesh_canAccess == null) {
                 Mesh_canAccess = typeof(Mesh).GetProperty("canAccess", System.Reflection.BindingFlags.NonPublic | System.Reflection.BindingFlags.Instance).GetMethod;
-
-            if(Mesh_canAccess != null)
-            {
+            }
+            if(Mesh_canAccess != null) {
                 try {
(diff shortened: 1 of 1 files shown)


package/com.unity.formats.usd/Runtime/Scripts/IO/Geometry/MeshExporter.cs
```diff
@@ -176,8 +176,8 @@ public static class MeshExporter {
         Debug.LogWarning("Null mesh for: " + path);
         return;
       }
-      if (mesh.isReadable == false) {
-        Debug.LogWarning("Mesh not readable: " + objContext.path);
+      if (mesh.isReadable == false && !Application.isEditor) {
```
jcowles: Are you sure this is true? I believe even in-editor meshes can be non-readable at run-time.

jcowles:
Oh also, please rebase against `dev` instead of `master`

package/com.unity.formats.usd/Runtime/Scripts/IO/Geometry/MeshExporter.cs
soraryu: Seems we are both right and there's actually 3 cases for the isReadable bool. Fancy quantum logic I guess.
(See https://docs.unity3d.com/ScriptReference/Mesh-isReadable.html).

1. it always works in Edit Mode
2. it works in Play Mode when **not** accessed from inside the game loop (accessed from Editor works)
3. it does not work in Play Mode when accessed from inside the game loop (this matches behaviour in build).

I updated the code to properly check for that. Unfortunately the "canAccess" property on a mesh, which controls that behaviour, is internal`, so I had to use reflection - not ideal.
A better check would be some kind of "Application.isInsideGameLoop". Are you aware of something like that?

soraryu:
I rebased on dev, however please see my comment above. Would be great if we can get rid of the reflection again (but it doesn't really hurt inside the #if UNITY_EDITOR).

package/com.unity.formats.usd/Runtime/Scripts/IO/Geometry/MeshExporter.cs
jcowles: Wow, interesting. I don't know off hand, but I'll ask around.

soraryu:
I added better log formatting and also pinging of objects causing trouble.
Additionally, I added more checks around the reflection access to make sure we're not messing things up if the internal API changes.

The problem with just "not supporting it" is that this would prevent all such exports from Editor in play mode, not just the ones through game loop. At least to me that would not be acceptable; my main workflow for exporting all kinds of things from Unity (GLTF/FBX/USDZ) is exporting from specific states during play mode. Setting the readable flags shouldn't be necessary, UX-wise, if just exporting from Editor. I added another warning log that the export would fail in a Build and only works in Editor.

soraryu:
OK, done

jcowles:
Awesome, thank you!!

PR #124, opened 2019-08-10
Better handling of non-matching submesh indices + materials

soraryu:
Check that materials are not null, change LogError to LogWarning (this is not corrupting data).
Also, better logs for what went wrong


soraryu:
Sure, just rebased on dev

jcowles:
Awesome, thanks!

PR #127, opened 2019-08-22
Linux Standalone Fixes (FTV-274)

cpredmann:


package/com.unity.formats.usd/Editor/Scripts/UsdBuildPostProcess.cs
```diff
@@ -24,6 +24,13 @@ public class UsdBuildPostProcess {
 #if UNITY_EDITOR_OSX
```
clusty: Nitpicking:
at least inside unity trunk, ifdefs are discouraged. I would make a switch/case since target tells you what platform are you building.
If not, at at the very least get rid of **ELSE**( if you build a switch/PS4 or whatever strange game, you execute the **ELSE** part)

clusty: Thinking more about this, this is VERY WRONG :)
You can build multiple platforms from one editor. UNITY_EDITOR_* is not the correct mechanism.

cpredmann: I'm fine with fixing it on this branch (and completely understand your reasoning), I was just sticking with the convention that's already in place, adding in the elif.

jcowles:
+1, great catch @clusty and thanks for the fix @cpredmann !

jcowles:
BTW, we have no Linux setup in the Yamato/CI build -- but maybe that's a separate issue?

PR #128, opened 2019-09-06
Add support for Sphere schema

lucillecaillaud:
99% of the changes in this PR are related to the work done when investigating the SideFX problem with importing crowd agent (USDSkel stuff). It consists of:

* Adding support to import Sphere schema (`SphereSample`, `SphereImporter`)
* Handling skinned primitive that inherits skeleton data from parents but don't define those themselves (`SkeletonImporter`)
* Intruder change: Make the `ImporterMesh` Sample works in edit mode.

I'm not confident with all the changes so be critical!


package/com.unity.formats.usd/Runtime/Scripts/IO/Scene/SceneImporter.cs
```diff
@@ -518,7 +518,16 @@ public ImportException(string message, System.Exception innerException)
             GameObject go = primMap[pathAndSample.path];
             NativeImporter.ImportObject(scene, go, scene.GetPrimAtPath(pathAndSample.path), importOptions);
             XformImporter.BuildXform(pathAndSample.path, pathAndSample.sample, go, importOptions, scene);
-            CubeImporter.BuildCube(pathAndSample.sample, go, importOptions);
+
+            pxr.UsdSkelSkinningQuery skinningQuery = primMap.SkelCache.GetSkinningQuery(scene.GetPrimAtPath(pathAndSample.path));
+            primMap.SkinningQueries[pathAndSample.path] = skinningQuery;
+            bool skinnedMesh =
+              importOptions.importSkinning
+              && primMap.SkelCache != null
```
jcowles: If SkelCache is null, an error would have been thrown above when calling primMap.SkelCache.GetSkinningQuery.

package/com.unity.formats.usd/Runtime/Scripts/IO/Scene/SceneImporter.cs
```diff
@@ -518,7 +518,16 @@ public ImportException(string message, System.Exception innerException)
...
+              && skinningQuery.IsValid();
+
+            CubeImporter.BuildCube(pathAndSample.sample, go, importOptions, skinnedMesh);
```
jcowles: Really nice how you boiled this down to a simple bool flag

package/com.unity.formats.usd/Runtime/Scripts/IO/Scene/SceneImporter.cs
```diff
@@ -527,6 +536,35 @@ public ImportException(string message, System.Exception innerException)
...
+          {
+            GameObject go = primMap[pathAndSample.path];
+            NativeImporter.ImportObject(scene, go, scene.GetPrimAtPath(pathAndSample.path), importOptions);
+            XformImporter.BuildXform(pathAndSample.path, pathAndSample.sample, go, importOptions, scene);
+
+            pxr.UsdSkelSkinningQuery skinningQuery = primMap.SkelCache.GetSkinningQuery(scene.GetPrimAtPath(pathAndSample.path));
+            primMap.SkinningQueries[pathAndSample.path] = skinningQuery;
+            bool skinnedMesh =
+              importOptions.importSkinning
+              && primMap.SkelCache != null
```
jcowles: Same issue here, if SkelCache is null, an error would have been trhown

package/com.unity.formats.usd/Runtime/Scripts/IO/Skel/SkeletonImporter.cs
```diff
@@ -106,6 +106,10 @@ public static class SkeletonImporter {
       }
     }
 
+    /// <summary>
+    /// Unity expect bind transforms to be the bone's inverse transformation matrix.
+    /// USD don't do that, so this function do it for us, and so prep the data to be used in Unity.
+    /// <summary>
```
jcowles: English bugs :)

/// Unity **expects** bind transforms to be the bone's inverse transformation matrix.
/// USD *doesn't* do that, so this function **does** it for us, **prepping** the data to be used in Unity.

src/USD.NET/serialization/UsdIo.cs
```diff
@@ -203,7 +203,7 @@ public class UsdIo {
 
       PropertyInfo[] properties = Reflect.GetCachedProperties(t.GetType());
       FieldInfo[] fields = Reflect.GetCachedFields(t.GetType());
-      object value = t;
+      object value = t;  // TODO: is this variable useful? Doesn't seem to be used. See line 245 bellow as well.
```
jcowles: I don't know what my crazy logic was here, I would try removing this and running the Tests project in the USD.NET solution.

package/com.unity.formats.usd/Runtime/Scripts/IO/Skel/SkeletonImporter.cs
```diff
@@ -134,15 +138,20 @@ public static class SkeletonImporter {
...
                                         string skelPath,
                                         SkeletonSample skeleton,
-                                        SkelBindingSample meshBinding,
+                                        UsdSkelSkinningQuery skinningQuery,
                                         GameObject go,
                                         PrimMap primMap,
                                         SceneImportOptions options) {
-      string[] joints = meshBinding.joints;
+      VtTokenArray jointsAttr = new VtTokenArray();
+      skinningQuery.GetJointOrder(jointsAttr);
```
jcowles: I believe this function can fail -- it might be worth checking the API to see if it returns a bool to indicate success.

package/com.unity.formats.usd/Runtime/Scripts/IO/Skel/SkeletonImporter.cs
```diff
@@ -134,15 +138,20 @@ public static class SkeletonImporter {
...
+      string[] joints = IntrinsicTypeConverter.FromVtArray(jointsAttr);
+
+      UsdGeomPrimvar jointWeights = skinningQuery.GetJointWeightsPrimvar();
+      UsdGeomPrimvar jointIndices = skinningQuery.GetJointIndicesPrimvar();
```
jcowles: Both of these objects may be invalid

package/com.unity.formats.usd/Runtime/Scripts/IO/Skel/SkeletonImporter.cs
```diff
@@ -160,7 +169,7 @@ public static class SkeletonImporter {
       }
 
       var mesh = smr.sharedMesh;
-      var geomXf = meshBinding.geomBindTransform.value;
+      Matrix4x4 geomXf = UnityTypeConverter.FromMatrix(skinningQuery.GetGeomBindTransform());
```
jcowles: Does the time argument matter here? I *think* the bind pose can be animated, but maybe we don't support that?

package/com.unity.formats.usd/Runtime/Scripts/IO/Skel/SkeletonImporter.cs
```diff
@@ -224,39 +233,42 @@ public static class SkeletonImporter {
       }
       smr.bones = bones;
 
-      int[] indices = meshBinding.jointIndices.value;
-      float[] weights = meshBinding.jointWeights.value;
+      int[] indices = IntrinsicTypeConverter.FromVtArray((VtIntArray)jointIndices.GetAttr().Get());
+      float[] weights = IntrinsicTypeConverter.FromVtArray((VtFloatArray)jointWeights.GetAttr().Get());
```
jcowles: `GetAttr()` can return an invalid UsdAttribute, which might segfault (I think)
`Get()` can also fail and return an empty VtValue, which I think will then fail to cast

Again are these time varying? If so and we do not yet support that, it might be good to add a comment here.

package/com.unity.formats.usd/Runtime/Scripts/IO/Geometry/CubeImporter.cs
```diff
@@ -27,17 +27,19 @@ public static class CubeImporter {
...
       Material mat = null;
 
       var cubeGo = GameObject.CreatePrimitive(PrimitiveType.Cube);
       var unityMesh = cubeGo.GetComponent<MeshFilter>().sharedMesh;
       GameObject.DestroyImmediate(cubeGo);
 
+      // Because Unity only handle a cube with a default size, the custom size of it is define by the localScale
+      // transform. This also need to be taken into account while computing the Unity extent of the mesh (see bellow).
+      float size = (float)usdCube.size;
+      go.transform.localScale = go.transform.localScale * size;
```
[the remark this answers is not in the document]
lucillecaillaud: If I'm not mistaken, any prim are treated as Xformable before they are treated as mesh (Mesh, Sphere, Cube). So I assumed that go.transform already contains any xformable information. According to my few test it seems to be the case.

package/com.unity.formats.usd/Runtime/Scripts/IO/Geometry/CubeImporter.cs
```diff
@@ -201,16 +206,33 @@ public static class CubeImporter {
         mat = options.materialMap.InstantiateSolidColor(Color.white);
       }
 
+      // Create Unity mesh.
+      Renderer renderer;
+      if (skinnedMesh)
+      {
+        renderer = ImporterBase.GetOrAddComponent<SkinnedMeshRenderer>(go);
+
+        if (((SkinnedMeshRenderer)renderer).sharedMesh == null)
```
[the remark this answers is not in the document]
lucillecaillaud: Right!
And also, do you have an idea on how I could get rid of the duplication of this bit that is in both `if` and `else` block?
```
if (skinnedRenderer.sharedMesh == null) {
  skinnedMeshRenderer = Mesh.Instantiate(unityMesh);
}
```

package/com.unity.formats.usd/Runtime/Scripts/IO/Geometry/SphereImporter.cs
```diff
@@ -0,0 +1,127 @@
...
+      }
+
+      if (usdSphere.colors != null && ShouldImport(options.meshOptions.color)) {
+        // NOTE: The following color conversion assumes PlayerSettings.ColorSpace == Linear.
+        // For best performance, convert color space to linear off-line and skip conversion.
+
+        if (usdSphere.colors.Length == 1) {
+          // Constant color can just be set on the material.
+          mat = options.materialMap.InstantiateSolidColor(usdSphere.colors[0].gamma);
+          Debug.Log("constant colors assigned");
```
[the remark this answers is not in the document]
lucillecaillaud: Oups, good catch

package/com.unity.formats.usd/Runtime/Scripts/IO/Geometry/SphereImporter.cs
```diff
@@ -0,0 +1,127 @@
...
+        {
+          Debug.LogWarning(
+            "Only constant color are supported for sphere (can't handle "
+            + usdSphere.colors.Length
+            + " color values)"
+          );
+        }
+      }
+
+      if (mat == null) {
```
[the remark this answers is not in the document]
lucillecaillaud: Yes I agree, I just don't really know yet what is really interesting to refactor or not and how. For now I'm using the technic of "make it work" that allow me to discover the code more and more, with the hope that at some point I'll have an idea of how to "cleverly" refactor stuff.

package/com.unity.formats.usd/Runtime/Scripts/IO/Scene/SceneImporter.cs
lucillecaillaud: Right, changing that! And actually putting the computation of `skinnedMesh` in a function.

package/com.unity.formats.usd/Runtime/Scripts/IO/Geometry/CubeImporter.cs
```diff
@@ -27,17 +27,19 @@ public static class CubeImporter {
     /// </summary>
     public static void BuildCube(CubeSample usdCube,
                                  GameObject go,
-                                 SceneImportOptions options) {
-
-
-      var mf = ImporterBase.GetOrAddComponent<MeshFilter>(go);
-      var mr = ImporterBase.GetOrAddComponent<MeshRenderer>(go);
+                                 SceneImportOptions options,
+                                 bool skinnedMesh=false) {
```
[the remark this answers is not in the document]
jcowles: Typo here: Wheither -> Whether

package/com.unity.formats.usd/Runtime/Scripts/IO/Geometry/SphereImporter.cs
```diff
@@ -0,0 +1,127 @@
...
+
+namespace Unity.Formats.USD {
+
+  /// <summary>
+  /// A collection of methods used for importing USD Sphere data into Unity.
+  /// </summary>
+  public static class SphereImporter {
+
+    /// <summary>
+    /// Copy sphere data from USD to Unity with the given import options.
```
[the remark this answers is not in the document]
jcowles: Typo: Wheither -> Whether

package/com.unity.formats.usd/Runtime/Scripts/IO/Skel/SkeletonImporter.cs
jcowles: Does UsdGeomPrimvar have an `IsValid()` function? If so, this would be better than `IsDefined`, though IsDefined /may/ still do what you want (not 100% sure)

package/com.unity.formats.usd/Runtime/Scripts/IO/Skel/SkeletonImporter.cs
jcowles: "Bon weights" -> "Bone weights"

Btw, if the `GetAttr()` call returns an invalid attribute, I think it may segfault. So you want something like this:

```
UsdAttribute jointIndicesAttr = jointIndices.GetAttr()
if (jointIndicesAttr.IsValid()) {
   VtValue indicesValue = jointIndicesAttr.Get();
   if (!indicesValue.IsEmpty()) {
     indices = IntrinsicTypeConverter.FromVtArray((VtIntArray)indicesValue)
   }
}
```

lucillecaillaud: But I can't see any case where it would return an invalid attribute. Do you have one in mind?

package/com.unity.formats.usd/Runtime/Scripts/IO/Skel/SkeletonImporter.cs
lucillecaillaud: There is no `IsValid` function, I tested `IsDefined` in various cases, it returns `false` when the attribute doesn't exist, return `true` otherwise. I also tried to set the attribute but with no values, in this case `IsDefine` still return `true`, and all `GetAttr`, `GetInterpolation`, etc. will return default values.

jcowles:
LGTM!

package/com.unity.formats.usd/Runtime/Scripts/IO/Geometry/CubeImporter.cs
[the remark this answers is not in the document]
jcowles: Not without complicating things... :/

PR #129, opened 2019-09-17
Standardize handedness transform to match other file format (FTV-299)

lucillecaillaud:
FBX, Alembic, Obj (and probably other formats) change their handedness to match the left hand coordinate system of Unity by inverting the X vector.

It's not conventional to do that but USD still has to be able to match that. One option is added to the Change Handedness list, "SlowAndSafe_asFBX" that allow to apply the same change of basis as for FBX.

By default, USD will still change basis allong the Z axis and not X.


package/com.unity.formats.usd/Runtime/Scripts/IO/Scene/SceneImporter.cs
```diff
@@ -252,6 +252,20 @@ public ImportException(string message, System.Exception innerException)
         throw new ImportException("Null USD Scene");
       }
 
+      // The matrix to change the handedness (the basis) is different for FBX and doesn't
+      // follow convention. This bit change this basisChange matrix to match the user options.
+      // [DEFAULT] SlowAndSafe is the standard and transform (X,Y,Z) basis to (X,Y,-Z)
+      // SlowAndSafe_AsFBX is consistent with FBX/Alembic/Obj ans transform (X,Y,Z) basis to (-X,Y,Z)
```
jcowles: Maybe:
```
The matrix to convert USD (right-handed) to Unity (left-handed) is different for the legacy FBX importer
and incorrectly swaps the X-axis rather than the Z-axis. This changes the basisChange matrix to match the
user options. See SceneImportOptions.BasisTransformation for additional details.
```

I think there is some way to literally reference that class (BasisTransformation) from the doc string above, but I forget how. It's like \<ref\> or something.

src/USD.NET.Unity/UnityTypeConverter.cs
```diff
@@ -28,23 +35,18 @@ public class UnityTypeConverter : IntrinsicTypeConverter {
...
       // TODO(jcowles): the change of basis matrix should probably be cached.
-      var basisChange = UnityEngine.Matrix4x4.identity;
-      // Invert the forward vector.
-      basisChange[2, 2] = -1;
 
       // Note that the fully general solution is basisChange * m * basisChange.inverse, however
       // basisChange and basisChange.inverse are identical. Just aliasing here so the math below
       // reads correctly.
-      var basisChangeInverse = basisChange;
+      var basisChangeInverse = UnityTypeConverter.basisChange;
```
jcowles: It is no longer guaranteed that the inverse is the same as the basisChange now -- not sure if you don't care about that (just leave it as-is) or want to expose basisChangeInverse as well...

lucillecaillaud: I computed the inverse of the matrix with -1 in [0, 0] and it gave me the same matrix.

jcowles: Yes, for your example they are the same, but now that matrix can be set to anything, there is no guarantee that the it is involutory, or even invertable at all.

lucillecaillaud: Right... but do we really allow any matrix?... it's true that I don't do any verification :/
Should I add a check or at least a comment/warning above?

jcowles: Well, we don't support any matrix, but we allow it because the matrix is public.

You could just expose the inverse matrix also and leave it to the user to set both correctly. That would also have the nice side effect of caching the inverse matrix, instead of computing it every time.

BTW, this matrix probably should be initialized to the default value, otherwise it only has the correct default when calling APIs via SceneImporter.

PR #130, opened 2019-09-26
Reduce the size of the float comparison epsilon

Secticide:
Fixes issues with meshes scaled to low values.


jcowles:
We could probably do something a little more rigorous, but this is definitely an improvement over the previous massive epsilon - looks good to me!

PR #137, opened 2019-10-24
Fix MeshImporter to handle attributes that are faceVarying correctly

lucillecaillaud:
* If any attributes is faceVarying: "unroll" the vertices to be the same size of the indices.
* Attributes that are faceVarying doesn't need to be unrolled anymore
* Attributes that are uniform are handled like if they were faceVarying
* MeshSampleBase now have Primvar to store uv values (it should actually be the case also for tangents and colors)


package/com.unity.formats.usd/Runtime/Scripts/IO/Geometry/MeshImporter.cs
```diff
@@ -603,55 +676,120 @@ public class GeometrySubsets {
...
     }
 
-    static void LoadPrimvars(Scene scene,
-                             Mesh unityMesh,
-                             string usdMeshPath,
-                             List<string> primvars,
-                             int[] faceVertexCounts,
-                             int[] faceVertexIndices) {
-      if (primvars == null || primvars.Count == 0) { return; }
+    private static Vector3[] SanitizePointsForUnity(Vector3[] vertices, int[] indices)
```
jcowles: Name game: this is not really "sanitizing" points, it's really expanding the indices (or unrolling them?).

package/com.unity.formats.usd/Runtime/Scripts/IO/Geometry/MeshImporter.cs
```diff
@@ -266,67 +268,84 @@ public class GeometrySubsets {
...
         Profiler.BeginSample("Triangulate Mesh");
-        if (options.meshOptions.triangulateMesh) {
-          // Triangulate n-gons.
-          // For best performance, triangulate off-line and skip conversion.
-          if (usdMesh.faceVertexIndices == null) {
+        if (options.meshOptions.triangulateMesh)
+        {
+          // TODO: I think that vertex indices and counts are mandatory attributes for a prim with a Mesh schema,
+          //       we could have a generic validation phase for that (generic = uniform checks of mandatory
+          //       attributes for any schema).
```
jcowles: Broken assets often are missing data that should be required :/

package/com.unity.formats.usd/Runtime/Scripts/IO/Geometry/MeshImporter.cs
```diff
@@ -337,64 +356,78 @@ public class GeometrySubsets {
           // TODO: this should be moved to C++
           int[] offsetMapping = new int[usdMesh.faceVertexCounts.Length];
           int curOffset = 0;
-          for (int i = 0; i < usdMesh.faceVertexCounts.Length; i++) {
+          for (int i = 0; i < usdMesh.faceVertexCounts.Length; i++)
+          {
```
jcowles: Curly braces on new line is actually the Unity style, but I would make that update in a separate change (across the entire code base), since it makes it harder to see the diff in this PR

package/com.unity.formats.usd/Runtime/Scripts/IO/Geometry/MeshImporter.cs
```diff
@@ -266,67 +268,84 @@ public class GeometrySubsets {
...
+          newVertices = new Vector3[usdMesh.points.Length];
+          for (int i = 0; i < usdMesh.points.Length; i++)
+          {
+            newVertices[i] = UnityTypeConverter.ChangeBasis(usdMesh.points[i]);
           }
         }
+        
+        if (IsThereFaceVaryingAttributes(usdMesh, options))
+        {
+          newVertices = SanitizePointsForUnity(newVertices ?? usdMesh.points, usdMesh.faceVertexIndices);
```
jcowles: I would add a comment here explaining what's happening and how it affects the rest of the import process below.

jcowles: Also, the faces only need to be split on seams, thought that is much more of a pain to implement.

package/com.unity.formats.usd/Runtime/Scripts/IO/Geometry/MeshImporter.cs
```diff
@@ -337,64 +356,78 @@ public class GeometrySubsets {
...
+          foreach (var nameAndSubset in geomSubsets.Subsets)
+          {
             var newFaceIndices = new List<int>();
-            foreach (var faceIndex in nameAndSubset.Value) {
+            foreach (var faceIndex in nameAndSubset.Value)
+            {
               newFaceIndices.Add(faceIndex + offsetMapping[faceIndex]);
-              for (int i = 1; i < usdMesh.faceVertexCounts[faceIndex] - 2; i++) {
+              for (int i = 1; i < usdMesh.faceVertexCounts[faceIndex] - 2; i++)
+              {
```
jcowles: You should definitely test an asset that has geometry subsets.

package/com.unity.formats.usd/Runtime/Scripts/IO/Geometry/MeshImporter.cs
```diff
@@ -422,116 +455,143 @@ public class GeometrySubsets {
...
           }
-        } else if (usdMesh.colors.Length == usdMesh.points.Length) {
-          // Vertex colors map on to verts.
-          // TODO: move the conversion to C++ and use the color management API.
-          for (int i = 0; i < usdMesh.colors.Length; i++) {
-            usdMesh.colors[i] = usdMesh.colors[i];
+        }
+        else
+        {
+          Color[] sanitizedColors;
```
jcowles: Name game again: "sanitized" is a little misleading

package/com.unity.formats.usd/Runtime/Scripts/IO/Geometry/MeshImporter.cs
```diff
@@ -603,55 +676,120 @@ public class GeometrySubsets {
...
+      return sanitizedPoints;
+    }
+
+    private static bool IsThereFaceVaryingAttributes(MeshSample usdMesh, SceneImportOptions options)
+    {
+      // If an attribute is face varying, its number of element will be usdMesh.indices.Length.
+      int pointCount = usdMesh.points.Length;
+      int faceCount = usdMesh.faceVertexCounts.Length;
+
+      // Potential face varying attributes are: normals, tangents, color and uv (st, uv, uv2. uv3. uv4).
```
jcowles: Any primvar requested by UsdPreviewSurface also needs to be included....

package/com.unity.formats.usd/Runtime/Scripts/IO/Geometry/MeshImporter.cs
```diff
@@ -603,55 +676,120 @@ public class GeometrySubsets {
...
+        }
 
         // Read the raw values.
         VtValue val = attr.Get(0.0);
-        if (val.IsEmpty()) { continue; }
+        if (val.IsEmpty())
+        {
+          continue;
+        }
         VtVec2fArray vec2fArray = UsdCs.VtValueToVtVec2fArray(val);
```
jcowles: Oof, I just noticed that we are assuming the primvar is a vec2 here...

package/com.unity.formats.usd/Runtime/Scripts/IO/Geometry/MeshImporter.cs
```diff
@@ -740,36 +909,41 @@ public class GeometrySubsets {
...
         uvVec = newUvs;
       }
 
       // If there are more UVs than verts, the UVs must be face varying, e.g. each vertex for
       // each face has a unique UV value. These values must be collapsed such that verts shared
       // between faces also share a single UV value.
-      if (uvVec.Length > vertexCount) {
-        uvVec = UnrollFaceVarying(vertexCount, uvVec, faceVertexCounts, faceVertexIndices);
+      if (uvVec.Length != vertexCount) {
+//        uvVec = UnrollFaceVarying(vertexCount, uvVec, faceVertexCounts, faceVertexIndices);
```
jcowles: dead comment?

jcowles (commented):
Incredible!!
This looks correct to me, though I'm not 100% sure about geometry subsets.

src/USD.NET.Unity/Geometry/MeshSampleBase.cs
```diff
@@ -37,19 +37,10 @@ public class MeshSampleBase : PointBasedSample {
...
-    // These are Unity friendly UV sets.
-    [VertexData] public object uv;
-    [VertexData] public object uv2;
-    [VertexData] public object uv3;
-    [VertexData] public object uv4;
+    public Primvar<object> st = new Primvar<object>();
+    public Primvar<object> uv = new Primvar<object>();
+    public Primvar<object> uv2 = new Primvar<object>();
+    public Primvar<object> uv3 = new Primvar<object>();
+    public Primvar<object> uv4 = new Primvar<object>();
```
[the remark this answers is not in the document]
lucillecaillaud: Arf... true... Is there a way to keep the backward compatibility?

Or, simply release a minor (or, actually a major) and disclaim that it's not backward compatible with existing script that use that? And, don't group the release of this feature with the one already merged.

package/com.unity.formats.usd/Runtime/Scripts/IO/Geometry/MeshImporter.cs
lucillecaillaud: Yes, but what I mean is that maybe... we could have a function that given a prim and it's schema, validate that all mandatory attributes are valid, and if not, ignore the prim... or something like that... and avoid too much "if blabla != null" in the code.

package/com.unity.formats.usd/Runtime/Scripts/IO/Geometry/MeshImporter.cs
lucillecaillaud: I know, I'm sorry, I did it for readability when I worked on it, but I should have removed it before doing the PR... I was lazy... I'll do it in my next commit :)

package/com.unity.formats.usd/Runtime/Scripts/IO/Geometry/MeshImporter.cs
lucillecaillaud: Do we have a known one?

package/com.unity.formats.usd/Runtime/Scripts/IO/Geometry/MeshImporter.cs
lucillecaillaud: The `|| options.ShouldBindMaterials` is here for that... It is a dirty shortcut... but I'm not sure how to check all the primvar requested by UsdPreviewSurface?

package/com.unity.formats.usd/Runtime/Scripts/IO/Geometry/MeshImporter.cs
lucillecaillaud: Should we change that? How?

package/com.unity.formats.usd/Runtime/Scripts/IO/Geometry/MeshImporter.cs
lucillecaillaud: Actually I'm missing some refactoring here as well... I did nothing except commenting the call to the function ^^

package/com.unity.formats.usd/Runtime/Scripts/IO/Skel/SkeletonImporter.cs
```diff
@@ -274,13 +275,35 @@ public static class SkeletonImporter {
 
       // Unity 2019 supports many-bone rigs, older versions of Unity only support four bones.
 #if UNITY_2019
+      if (!isConstant && mesh.vertexCount == faceVertexIndices.Length)  // vertices were unrolled
+      {
+        int[] newIndices = new int[faceVertexIndices.Length * indicesElementSize];
+        float[] newWeights = new float[faceVertexIndices.Length * weightsElementSize];
+        for (int i = 0; i < faceVertexIndices.Length; i++)
+        {
+          for (int idxSize = 0; idxSize < indicesElementSize; idxSize++)
```
judubu: You can merge both loops together as indicesElementSize always equals weightsElementSize

judubu:
The importer is being rewritten.

PR #139, opened 2019-10-29
Prep for release of 1.0.2

lucillecaillaud:
- Update changelog with added/changed/removed features
- Update package version to 1.0.2-preview.1
- Update sample to point to the right package version
- Update the upm-ci command to run test coverage

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -51,6 +51,9 @@ packages.config
 # OSX
 .DS_Store
 
+# UPM
+upm-ci~
+
 TestProject/Library
 TestProject/ProjectSettings
 TestProject/Temp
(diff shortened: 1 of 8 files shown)


jcowles:
@soraryu FYI

soraryu:
I know you're probably aware of this, but given that Unity officially recommends and for some platforms requires using IL2CPP for Builds (for speedup, for security, for Android x64 support, for WebGL, ...) I think #100 should really be solved before releasing another non-preview version and then having many months of wait time before the next release.

lucillecaillaud:
@soraryu, this is a preview release, there is no "non-preview" release scheduled for now. This package needs work to be "production ready" and we're currently planning and working on the steps toward that. IL2CPP will definitly be taken into account in those plan.

PR #148, opened 2020-01-10
Add yamato nightly jobs

judubu:
I kept the 2 existing jobs and added nightly jobs for master and dev.
I guess there's no other way to test it than wait for the jobs to run?

### DIFF
--- a/.yamato/upm-ci.yml
+++ b/.yamato/upm-ci.yml
@@ -20,18 +20,18 @@ test_platforms:
     flavor: m1.mac
 ---
 pack:
- name: Pack
- agent:
-   type: Unity::VM::osx
-   image: buildfarm/mac:stable
-   flavor: m1.mac
- commands:
-   - npm install upm-ci-utils@stable -g --registry https://api.bintray.com/npm/unity/unity-npm
-   - upm-ci package pack --package-path package/com.unity.formats.usd
- artifacts:
-   packages:
-     paths:
-       - "upm-ci~/packages/**/*"
+  name: Pack
+  agent:
+    type: Unity::VM::osx
+      image: buildfarm/mac:stable
+      flavor: m1.mac
+    commands:
+      - npm install upm-ci-utils@stable -g --registry https://api.bintray.com/npm/unity/unity-npm
+      - upm-ci package pack --package-path package/com.unity.formats.usd
+    artifacts:
+      packages:
+        paths:
+          - "upm-ci~/packages/**/*"
 
 {% for editor in test_editors %}
 {% for platform in test_platforms %}
@@ -51,8 +51,8 @@ test_{{ platform.name }}_{{ editor.version }}:
         - "upm-ci~/test-results/**/*"
   dependencies:
     - .yamato/upm-ci.yml#pack
-{% endfor %}
-{% endfor %}
+      {% endfor %}
+      {% endfor %}
 
 # Don't run test coverage on Unity version older than 2019.2
 {% for editor in test_editors_old %}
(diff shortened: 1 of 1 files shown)


judubu:
That's actually what I'm used to as well but this repo didn't seem to follow the rule. I'll update the commit message and name things properly from now on.
Thanks

.yamato/upm-ci.yml
```diff
@@ -51,8 +51,8 @@ test_{{ platform.name }}_{{ editor.version }}:
         - "upm-ci~/test-results/**/*"
   dependencies:
     - .yamato/upm-ci.yml#pack
-{% endfor %}
-{% endfor %}
+      {% endfor %}
+      {% endfor %}
```
lucillecaillaud: No need to move this one, also it's better if it's stay aligned with the "begin"for.

.yamato/upm-ci.yml
```diff
@@ -72,8 +72,8 @@ test_{{ platform.name }}_{{ editor.version }}:
...
```
lucillecaillaud: Same as above

.yamato/upm-ci.yml
```diff
@@ -98,12 +98,41 @@ test_trigger:
...
     - .yamato/upm-ci.yml#pack
-    {% for editor in test_editors_all %}
-    {% for platform in test_platforms %}
+      {% for editor in test_editors_all %}
+      {% for platform in test_platforms %}
     - .yamato/upm-ci.yml#test_{{platform.name}}_{{editor.version}}
-    {% endfor %}
-    {% endfor %}
+      {% endfor %}
+      {% endfor %}
```
lucillecaillaud: Minor formatting comment: if we're at fixing the format of this file I would follow the way it was made in "run_preview_verified_staging" job, i.e. keep the jinja template stuff with no indentation as it's something apart from the regular yaml file and so it has no reason to follow the same yaml indentation (otherwise we could argue that nested for loops should have different indentation, but in some case it would cause problem, blah blah blah). Also it will be less changes that are not related to the purpose of the PR.

But I'm aware that there is no known official convention (or maybe there is and I don't know it) and so that's a "user preference" thing.

.yamato/upm-ci.yml
```diff
@@ -98,12 +98,41 @@ test_trigger:
...
+    packages:
+      paths:
+        - "upm-ci~/packages/**/*"
+  dependencies:
+    - .yamato/upm-ci.yml#pack
+      {% for editor in test_editors_all %}
+      {% for platform in test_platforms %}
+    - .yamato/upm-ci.yml#test_{{platform.name}}_{{editor.version}}
+      {% endfor %}
+      {% endfor %}
```
lucillecaillaud: Same as above.

.yamato/upm-ci.yml
```diff
@@ -124,8 +153,8 @@ run_preview_verified_staging:
...
     - .yamato/upm-ci.yml#pack
-  {% for editor in test_editors_all %}
-  {% for platform in test_platforms %}
+      {% for editor in test_editors_all %}
+      {% for platform in test_platforms %}
     - .yamato/upm-ci.yml#test_{{ platform.name }}_{{ editor.version }}
-  {% endfor %}
-  {% endfor %}
+      {% endfor %}
+      {% endfor %}
```
lucillecaillaud: Same as above.

.yamato/upm-ci.yml
judubu: Yeah I was checking the one in the recorder repo and given there was 2 different styles in the yamato usd file I used the one from recorder.
After a chat with @clusty we agreed on aligning to either the - or the . of the line before so that it doesn't break the flow and looks weird when more complex liquid instructions with indentations are needed.

.yamato/upm-ci.yml
judubu: Oh you're right. I'll fix that.

PR #149, opened 2020-01-10
Fix FBX loading mode

judubu:
Took over this one to learn more about the details of the importer.

When SlowAndSafeAsFBX mode is selected:
* complete the handedness change for meshes (x to -x)
* rotate the camera to compensate for the buggy handedness conversion


package/com.unity.formats.usd/Runtime/Scripts/IO/Geometry/XformImporter.cs
```diff
@@ -75,6 +75,19 @@ public static class XformImporter {
...
+#if !UNITY_2020_1_OR_NEWER
+      if (options.changeHandedness == BasisTransformation.SlowAndSafeAsFBX)
+      {
+        var cam = go.GetComponent<Camera>();
+        if (cam != null)
+        {
+          go.transform.rotation *= Quaternion.Euler(180.0f, 0.0f, 180.0f);
+        }
+      }
+#endif
```
lucillecaillaud: Just to be sure I understand correctly... this only change any imported camera right? And the camera importer doesn't already apply the basis change or whatever transformation is needed?

package/com.unity.formats.usd/Runtime/Scripts/IO/Geometry/MeshImporter.cs
```diff
@@ -264,7 +264,8 @@ public class GeometrySubsets {
       // rely on the UnityEngine.Mesh API.
 
       Material mat = renderer.sharedMaterial;
-      bool changeHandedness = options.changeHandedness == BasisTransformation.SlowAndSafe;
+      bool changeHandedness = options.changeHandedness == BasisTransformation.SlowAndSafe ||
+                              options.changeHandedness == BasisTransformation.SlowAndSafeAsFBX;
```
lucillecaillaud: That is comforting ^^, thanks for the spot!

package/com.unity.formats.usd/Runtime/Scripts/IO/Geometry/XformImporter.cs
judubu: Yes it only fixes cameras. I initially did it in CameraImporter but it falls apart when the cam position is animated because only XFormImporter gets called.

jcowles: Why is the extra rotation needed?

If this (see link below) is not the right transform, should we not just update it to whatever is needed, rather than applying a second transform? Or maybe I'm not understanding the problem...

https://github.com/Unity-Technologies/usd-unity-sdk/blob/a3fec00982670f51f5277ec646a0f195fd86b97c/package/com.unity.formats.usd/Runtime/Scripts/IO/Scene/SceneImporter.cs#L261

package/com.unity.formats.usd/Runtime/Scripts/IO/Geometry/MeshImporter.cs
jcowles: +1, nice catch!

package/com.unity.formats.usd/Runtime/Scripts/IO/Geometry/XformImporter.cs
judubu: The basis change is right for the meshes but the cameras and lights get a special treatment in the FBX importer. I empirically got the rotation working (btw I will change it to 180 in Y) but could not find exactly what was happening in the FBX importer to justify it. 

If you check FBXImporter/UnityFBX.cpp you can see that first the rotations are inverted in Y and Z (l387 ExtractQuaternionFromFBXEulerOld) and then a rotation of 90 in Y is applied (l449 ConvertComponents). I'm stuck there.

jcowles: huh, ok!

PR #152, opened 2020-01-30
Usd 20.02 upgrade (FTV-231)

judubu:


third_party/include/pxr/usd/sdf/path.h
```diff
@@ -0,0 +1,1256 @@
...
+}
+
+/// Return an iterator to the element of [\a begin, \a end) that is the longest
+/// prefix of the given path (including the path itself), if there is such an
+/// element, otherwise \a end.  The input range must be ordered according to
+/// SdfPath::operator<.  If your range's iterators' value_types are not SdfPath,
+/// but you can obtain SdfPaths from them (e.g. vector<pair<SdfPath,
+/// X>>::iterator), you can pass a function to extract the path from the
+/// dereferenced iterator in \p getPath.
+#if 0 // USD.NET
```
judubu: Added that file just to Ifdef this part

jcowles:
Do you know why the CI tests are failing?

CMakeFiles/CMakeError.log
```diff
@@ -0,0 +1,92 @@
+Compiling the CXX compiler identification source file "CMakeCXXCompilerId.cpp" failed.
```
jcowles: Intended to commit this?

CMakeFiles/3.16.0-rc3/CMakeCXXCompiler.cmake
```diff
@@ -0,0 +1,88 @@
+set(CMAKE_CXX_COMPILER "C:/Program Files (x86)/Microsoft Visual Studio/2017/Enterprise/VC/Tools/MSVC/14.16.27023/bin/Hostx64/x64/cl.exe")
```
jcowles: Why so many CMake changes?

package/com.unity.formats.usd/Runtime/Plugins/x86_64/Linux.meta
```diff
@@ -1,8 +0,0 @@
-fileFormatVersion: 2
```
jcowles: Linux and OSX support seem to be removed, is that intentional?

package/com.unity.formats.usd/Runtime/Plugins/x86_64/plugInfo.json
```diff
@@ -1,3 +0,0 @@
-{
-    "Includes": [ "share/" ]
```
jcowles: Intentionally removed?

package/com.unity.formats.usd/Runtime/Plugins/x86_64/share/plugInfo.json.meta
```diff
@@ -1,5 +1,5 @@
 fileFormatVersion: 2
-guid: b65269c81f13f44c9920bf91d258b33b
+guid: 1e56bcd3fdf778349b3524cc6cc85793
```
jcowles: GUIDs are the fundamental file identifier in the asset database and can cause upgrade issues when they change but the file isn't actually changing -- any clue why all these GUIDs changed?

package/com.unity.formats.usd/Runtime/Plugins/x86_64/plugInfo.json.meta
```diff
@@ -1,9 +0,0 @@
-fileFormatVersion: 2
```
jcowles: Intentionally removed?

package/com.unity.formats.usd/Runtime/Plugins/x86_64/plugInfo.json.meta
```diff
@@ -1,9 +0,0 @@
...
-guid: 490ba57b3b3952d43bcbe916c3b56dc7
-timeCreated: 1521838944
```
jcowles: I don't see any changes to actual plugInfo.json files, which is surprising, they almost always have modifications after an upgrade.

package/com.unity.formats.usd/Runtime/Plugins/x86_64/share/usdAbc/resources/plugInfo.json
```diff
@@ -1,27 +0,0 @@
-{
```
jcowles: Was the alembic plugin moved/removed?

jcowles (commented):
This is awesome, nice work!

I don't have a ton of context, so take my questions with a grain of salt (maybe the answers are obvious)

third_party/include/pxr/usd/sdf/path.h
jcowles: Thank you for calling this out!

src/Swig/pxr/usd/usd/usdAttribute.i
```diff
@@ -57,4 +57,7 @@ typedef std::vector<UsdAttribute> UsdAttributeVector;
 %ignore UsdAttribute::GetTimeSamples(std::vector<double>* times) const;
 %ignore UsdAttribute::Get(VtValue* value) const;
 
+%ignore UsdAttribute::Set(const char* value) const;
+%ignore UsdAttribute::Set(const char* value, UsdTimeCode time = UsdTimeCode::Default()) const;
```
lucillecaillaud: Why did you need to ignore those two functions?

src/Swig/pxr/usd/usdRi/tokens_Tokens.i
```diff
@@ -39,11 +39,17 @@
   public static TfToken aovName = new TfToken("aovName");
```
lucillecaillaud: That's a side question, maybe more for @jcowles ... I'm wondering why we have any work/bind for the usdRi API as we don't use RenderMan at all?

src/USD.NET/USD.NET.csproj
```diff
@@ -256,6 +256,9 @@
     <Compile Include="generated\pxr\usd\sdf\SdfStringListOpVector.cs" />
```
lucillecaillaud: A quick "improvement" could be to only include something like "generated\pxr\**\*.cs" once instead of listing each csharp file generated. That would eventually avoid seeing this file in future diff.

src/UsdCs/usdDefines.h
```diff
@@ -57,5 +57,7 @@ PXR_NAMESPACE_USING_DIRECTIVE
 #include "pxr/base/gf/vec4h.h"
 #include "pxr/base/gf/vec4i.h"
 #include "pxr/usd/sdf/assetPath.h"
+#include "pxr/usd/sdf/timeCode.h"
 
-typedef VtArray<SdfAssetPath> SdfAssetPathArray;
\ No newline at end of file
+typedef VtArray<SdfAssetPath> SdfAssetPathArray;
+typedef VtArray<SdfTimeCode> SdfTimeCodeArray;
```
lucillecaillaud: Like this github feature that highlight missing new line at the end!

lucillecaillaud (commented):
Phew, that was tedious to review ^^. Thanks for the update! Here is some general comment/questions:

I'm not sure why some meta files changed and why they changed this way. Did you deleted them and so they were regenerated... that could explain the uuid changes I guess and some others. Any clue?

Also not sure why so much files were removed. I guess you analysed they were not needed? Do you have a specific explanation or just realize they're just "noise"?

Just to make sure I didn't miss anything important, there's only the "bind material" part that needed to change in the actual package code, right?

Also, next time would be nice to add a description to the review, just to help know what are the important part to focus on in the middle of the 400+ changed files.

lucillecaillaud:
@jcowles about Yamato, seems at least some tests are failing. But I can see various errors, one that looks easy to fix is something like "Failed to open: C:/l4Sgff/Assets\sceneFile.usda"... not sure it's the only one to check.

@judubu did you run the test that are in the visual solutions? I do think that the one failing are the one that run in unity though (cause I don't think Yamato run any other kind of tests), but still good to make sure that also the one in the source are running correctly.

CMakeFiles/3.16.0-rc3/CMakeCXXCompiler.cmake
judubu: That Cmake folder was clearly not supposed to be commited... 😞

src/USD.NET/USD.NET.csproj
judubu: Can you do that in VS? As soon as you add a folder all the filepaths are added to the csproj.

package/com.unity.formats.usd/Runtime/Plugins/x86_64/Linux.meta
judubu: Overlook on my part until I noticed the failed CI jobs (thanks @clusty) 
I'm setting up bokken jobs to build usd and the bindings on all 3 platforms. I'll update the PR asap.

package/com.unity.formats.usd/Runtime/Plugins/x86_64/share/plugInfo.json.meta
judubu: Yeah I removed the whole folder when rebuilding to make sure I didn't have old dependencies hiding somewhere. So all the meta files were regenerated...
I should update the install script to remove all libs, keep the meta files and copy the fresh libs. I could add the usd dependencies to that as well, because it's all manual at the moment.

@lucillecaillaud is your WIP build/install script taking care of that at the moment?

package/com.unity.formats.usd/Runtime/Plugins/x86_64/share/usdAbc/resources/plugInfo.json
judubu: Hmmm, I'll have to check my usd build. But it should not. 🤔

src/Swig/pxr/usd/usd/usdAttribute.i
judubu: Those 2 were added recently in usd as convenience overload and swig was complaining about ambiguous definition with
`
 UsdAttribute::Set(VtValue value) A
UsdAttribute::Set(VtValue value, UsdTimeCode time = UsdTimeCode::Default())
`

package/com.unity.formats.usd/Runtime/Plugins/x86_64/share/plugInfo.json.meta
lucillecaillaud: The new "binding project" I'm working on doesn't care of Unity... so I have no meta file and will never have. So unfortunatly, my script won't help you for that.

src/USD.NET/USD.NET.csproj
lucillecaillaud: I did that on my side, it worked well. Do what you think it's best and works. After a second thought, maybe we prefer to have this explicit diff of new files and the one removed even if it's a bit big.

jcowles:
I recently upgraded to 20.08, but I'd like to make sure I didn't lose anything good you had in this upgrade (I did look over it), if you get a chance Julien.

Aside from that, I think this PR is now deprecated

judubu:
Looks like everything here is in the new PR. Thanks @jcowles 
Closing.

PR #156, opened 2020-02-12
Fix: load scope as xform (FTV-417)

judubu:
Fix https://github.com/Unity-Technologies/usd-unity-sdk/issues/154

Scopes are not Xformable per se but can contain payload and hold variantSets. If we want to be able to set the variant before loading the payload we need to load the Scope as a xform.


thomas-tu (commented):
Should the `ScopeSample` array be injected in the PrimMap like the others results?

thomas-tu:
Additionally, could we maybe add tests to make sure we load things properly with different cases?

judubu:
@thomas-tu The only reason we'd want the Scope prims added to the primMap would be to run specific import/export logic on them. Given that Scopes are represented in Unity as Xforms and that  the exporter automatically exports the whole parent hierarchy for any valid prims I don't think we need to add specific ScopeImporter/Exporter for now. https://github.com/Unity-Technologies/usd-unity-sdk/blob/b28f0fee38a2a286e199d1f6e0d260789f475279/package/com.unity.formats.usd/Runtime/Scripts/IO/Scene/SceneExporter.cs#L635-L637
That will be required if/when we focus on proper round tripping of USD assets.

package/com.unity.formats.usd/Tests/Editor/EditorTests.cs
```diff
@@ -1,4 +1,7 @@
 using NUnit.Framework;
+using UnityEngine;
+using USD.NET;
+using System.IO;
```
jcowles: AMAZING, A TEST!

package/com.unity.formats.usd/Tests/Editor/EditorTests.cs
```diff
@@ -9,5 +12,35 @@ public void DummyTest()
...
+        private GameObject usdRoot;
+        private string filepath = "test_#154/test_collections.usda";
+        
+        [SetUp]
+        public void SetUp()
+        {
+            InitUsd.Initialize();
+            var stage = pxr.UsdStage.Open(Path.Combine(Application.dataPath, filepath), pxr.UsdStage.InitialLoadSet.LoadNone);
+            var scene = Scene.Open(stage);
+            usdRoot = USD.UsdMenu.ImportSceneAsGameObject(scene);
```
jcowles: Should probably make a SceneLoader utility class (either in this change or a future change when adding new tests), so all tests don't have to implement this logic

package/com.unity.formats.usd/Runtime/Scripts/IO/Scene/HierarchyBuilder.cs
```diff
@@ -151,6 +151,7 @@ public struct Query<T> : IQuery where T : SampleBase, new() {
       }
       if (options.importTransforms) {
         FindPathsJob.queries[7] = (FindPathsJob.IQuery)new FindPathsJob.Query<XformSample>();
+        FindPathsJob.queries[8] = (FindPathsJob.IQuery)new FindPathsJob.Query<ScopeSample>();
```
jcowles: We may need to import scopes regardless of transform import state... I don't know if there is a use case, but I could imagine making a hierarchy of scopes, disabling `importTransforms` and then being confused why my scopes are showing up.

package/com.unity.formats.usd/Tests/Editor/EditorTests.cs
```diff
@@ -9,5 +12,35 @@ public void DummyTest()
...
+            scene.Close();
+        }
+        
+        [Test]
+        public void TestScopeWithoutChildren()
+        {
+            var scopeWithChildren = usdRoot.transform.Find("TestComponent/geom");
+            Assert.IsNotNull(scopeWithChildren);
```
jcowles: Do we want to assert anything about the data on the scope? E.g. that it has a UsdPrim component or that the transform is identity?

jcowles (approved):
Nice, love that you added a real test!

Some minor comments, but mostly just for your consideration.

package/com.unity.formats.usd/Tests/Editor/EditorTests.cs
judubu: That's what I've been telling myself for each branch I've worked on so far... 
I need to rewrite the loading part anyway because assets are not found on Yamato. Adding them to the package and accessing them through guid seems to be the correct way.

package/com.unity.formats.usd/Tests/Editor/EditorTests.cs
```diff
@@ -9,5 +12,35 @@ public void DummyTest()
...
+        {
+            InitUsd.Initialize();
+            var stage = pxr.UsdStage.Open(Path.Combine(Application.dataPath, filepath), pxr.UsdStage.InitialLoadSet.LoadNone);
+            var scene = Scene.Open(stage);
+            usdRoot = USD.UsdMenu.ImportSceneAsGameObject(scene);
+            scene.Close();
+        }
+        
+        [Test]
+        public void TestScopeWithoutChildren()
```
thomas-tu: We should apply some rules on the test naming. You would like to be able what broke from a log without to have to open the code. My rule of thumb: `<LogicName/MethodName>_<Input>_<ExpectedOutput>`.

Names become long but this is necessary to prevent time consuming back and forth between Yamato and a local repository.

package/com.unity.formats.usd/Tests/Editor/EditorTests.cs
```diff
@@ -9,5 +12,35 @@ public void DummyTest()
...
+        {
+            var scopeWithChildren = usdRoot.transform.Find("TestComponent/geom");
+            Assert.IsNotNull(scopeWithChildren);
+        }
+        
+        [Test]
+        public void TestScopeWithChildren()
```
thomas-tu: See my comment about `TestScopeWithoutChildren()`:


> We should apply some rules on the test naming. You would like to be able what broke from a log without to have to open the code. My rule of thumb: <LogicName/MethodName>_<Input>_<ExpectedOutput>.
> 
> Names become long but this is necessary to prevent time consuming back and forth between Yamato and a local repository.

package/com.unity.formats.usd/Tests/Editor/EditorTests.cs
thomas-tu: +1. 

You can make a base framework to do that for you. That will probably become very redundant. Plus, you delegate the knowledge of "How do I load my stuff on Yamato" to one place, allowing people to focus on their tests rather than Yamato-specific rules.

package/com.unity.formats.usd/Tests/Editor/EditorTests.cs
thomas-tu: @jcowles, is it valuable for this test? Or should they be in separate ones?

thomas-tu (changes requested):
Nice! Looks good!

I left a comment about naming tests. I think that will be important in the future.
Also, do you expect to re-use the `.usda` files in differents tests? I usually classify my assets in folder that is also name after the tests main class for tracking purposes. In this case, they would have been put in `ScopeLoadingTests/`.

package/com.unity.formats.usd/Tests/Editor/EditorTests.cs
judubu: You're right, I'll do both and add a little bit more because I'm realizing the test doesn't do much. Looks like I lost some bits between stashes...

package/com.unity.formats.usd/Tests/Editor/EditorTests.cs
judubu: I'll create a separate task for that

thomas-tu (approved):

judubu:
The yamato failure on trunk comes from the dev branch. Investigation in progress. Meanwhile I'm merging this branch which has been approved for ages and had its trunk tests passing before.

PR #159, opened 2020-04-01
FTV-244 prefab reset on play mode

judubu:
Remove the "rebuildAll on clone" logic and add guid as mesh names. (FTV-244)

We used to force a rebuild of the hierarchy when a UsdAsset (game object)
was duplicated to avoid sharing meshes and materials.
This caused an issue when game objects created from usd prefabs in
PlayMode where flagged for rebuild, losing all local edits.

I decided to remove the "rebuildAll on clone" logic and keep meshes and
materials shared between copies of UsdAsset imported as game objects.
That's the behavior I expect when I duplicate a usd asset reading the same
location in a usd file. It's also safe to do for skinned meshes as skinning is applied
in a vertex shader and the underlying shared mesh is not deformed.

The only drawback is that it won't work on deformed meshes in the case
you change playback speed between the copies. But the solution is simple,
you just have to reload and rebuild the hierachy in the UsdAsset.

I also set the name of the meshes with a guid to make it obvious that
the meshes are shared.

### DIFF
--- a/TestProject/Packages/manifest.json
+++ b/TestProject/Packages/manifest.json
@@ -1,12 +1,9 @@
 {
   "dependencies": {
-    "com.unity.ads": "2.0.8",
-    "com.unity.analytics": "3.2.2",
-    "com.unity.collab-proxy": "1.2.15",
+    "com.unity.analytics": "3.3.5",
+    "com.unity.ext.nunit": "1.0.0",
     "com.unity.formats.usd": "file:../../package/com.unity.formats.usd/",
-    "com.unity.package-manager-ui": "2.0.3",
-    "com.unity.purchasing": "2.0.3",
-    "com.unity.textmeshpro": "1.3.0",
+    "com.unity.test-framework": "1.1.11",
     "com.unity.modules.ai": "1.0.0",
     "com.unity.modules.animation": "1.0.0",
     "com.unity.modules.assetbundle": "1.0.0",
@@ -38,8 +35,7 @@
     "com.unity.modules.wind": "1.0.0",
     "com.unity.modules.xr": "1.0.0"
   },
-  "testables": 
-  [
+  "testables": [
     "com.unity.formats.usd"
   ]
 }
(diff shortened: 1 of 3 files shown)


judubu:
Hold on, PR was created to merge to master.
Recreating...

TestProject/Packages/manifest.json
```diff
@@ -38,8 +35,7 @@
     "com.unity.modules.wind": "1.0.0",
     "com.unity.modules.xr": "1.0.0"
   },
-  "testables": 
-  [
+  "testables": [
```
jcowles, line 38: Package manifest edits intended for this change?

package/com.unity.formats.usd/Runtime/Scripts/IO/Geometry/MeshImporter.cs
```diff
@@ -225,7 +225,7 @@ public class GeometrySubsets {
                              SceneImportOptions options) {
       var smr = ImporterBase.GetOrAddComponent<SkinnedMeshRenderer>(go);
       if (smr.sharedMesh == null) {
-        smr.sharedMesh = new Mesh();
+        smr.sharedMesh = new Mesh {name = Guid.NewGuid().ToString()};
```
jcowles, line 228: I wonder if there's a more meaningful name we could give here (just thinking out loud, this is not a request to change it)

package/com.unity.formats.usd/Runtime/Scripts/IO/Geometry/MeshImporter.cs
```diff
@@ -241,8 +241,10 @@ public class GeometrySubsets {
                              SceneImportOptions options) {
       var mf = ImporterBase.GetOrAddComponent<MeshFilter>(go);
       var mr = ImporterBase.GetOrAddComponent<MeshRenderer>(go);
-      if (mf.sharedMesh == null) {
-        mf.sharedMesh = new Mesh();
+      if (mf.sharedMesh == null)
+      {
```
jcowles, line 245: We should probably format the entire code base to conform to Unity conventions, in the mean time we should keep it self-consistent (just noticing the brace placement).

package/com.unity.formats.usd/Runtime/Scripts/IO/Geometry/MeshImporter.cs
```diff
@@ -241,8 +241,10 @@ public class GeometrySubsets {
...
+        mf.sharedMesh = new Mesh {name = Guid.NewGuid().ToString()};
+        mf.sharedMesh.MarkDynamic();
```
jcowles, line 247: Ideally, this would only be set if the mesh were animated in the USD file, given how many static meshes a USD set can have.

If not for this change, maybe just add a TODO comment here?

PR #160, opened 2020-04-01
Remove the "rebuildAll on clone" logic and add guid as mesh names. (F…

judubu:
…TV-244)

We used to force a rebuild of the hierarchy when a UsdAsset (game object)
was duplicated to avoid sharing meshes and materials.
This caused an issue when game objects created from usd prefabs in
PlayMode where flagged for rebuild, losing all local edits.

I decided to remove the "rebuildAll on clone" logic and keep meshes and
materials shared between copies of UsdAsset imported as game objects.
That's the behavior I expect when I duplicate a usd asset reading the same
location in a usd file. It's also safe to do for skinned meshes as skinning is applied
in a vertex shader and the underlying shared mesh is not deformed.

The only drawback is that it won't work on deformed meshes in the case
you change playback speed between the copies. But the solution is simple,
you just have to reload and rebuild the hierachy in the UsdAsset.

I also set the name of the meshes with a guid to make it obvious that
the meshes are shared.

### DIFF
--- a/package/com.unity.formats.usd/Runtime/Scripts/IO/Geometry/MeshImporter.cs
+++ b/package/com.unity.formats.usd/Runtime/Scripts/IO/Geometry/MeshImporter.cs
@@ -225,7 +225,7 @@ namespace Unity.Formats.USD {
                              SceneImportOptions options) {
       var smr = ImporterBase.GetOrAddComponent<SkinnedMeshRenderer>(go);
       if (smr.sharedMesh == null) {
-        smr.sharedMesh = new Mesh {name = Guid.NewGuid().ToString()};
+        smr.sharedMesh = new Mesh {name = UniqueMeshName(go.name)};
       }
 
       BuildMesh_(path, usdMesh, smr.sharedMesh, geomSubsets, go, smr, options);
@@ -241,10 +241,8 @@ namespace Unity.Formats.USD {
                              SceneImportOptions options) {
       var mf = ImporterBase.GetOrAddComponent<MeshFilter>(go);
       var mr = ImporterBase.GetOrAddComponent<MeshRenderer>(go);
-      if (mf.sharedMesh == null)
-      {
-        mf.sharedMesh = new Mesh {name = Guid.NewGuid().ToString()};
-        mf.sharedMesh.MarkDynamic();
+      if (mf.sharedMesh == null) {
+        mf.sharedMesh = new Mesh {name = UniqueMeshName(go.name)};
       }
 
       BuildMesh_(path, usdMesh, mf.sharedMesh, geomSubsets, go, mr, options);
@@ -850,6 +848,14 @@ namespace Unity.Formats.USD {
     /// </summary>
     private static bool ShouldCompute(ImportMode mode) {
       return mode == ImportMode.Compute || mode == ImportMode.ImportOrCompute;
+    }
+
+    /// <summary>
+    /// Returns a unique mesh name by appending a short guid to the given string 
+    /// </summary>
+    private static string UniqueMeshName(string meshName) {
+      var shortGuid = Convert.ToBase64String(Guid.NewGuid().ToByteArray());
+      return meshName + "_" +  shortGuid.Substring(0, shortGuid.Length-2); 
     }
   }
 }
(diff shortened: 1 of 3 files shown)


package/com.unity.formats.usd/Runtime/Scripts/IO/Geometry/MeshImporter.cs
```diff
@@ -225,7 +225,7 @@ public class GeometrySubsets {
                              SceneImportOptions options) {
       var smr = ImporterBase.GetOrAddComponent<SkinnedMeshRenderer>(go);
       if (smr.sharedMesh == null) {
-        smr.sharedMesh = new Mesh();
```
jcowles, line 228: I wonder if there's a more meaningful name we could give here (just thinking out loud, this is not a request to change it)

package/com.unity.formats.usd/Runtime/Scripts/IO/Geometry/MeshImporter.cs
```diff
@@ -241,8 +241,10 @@ public class GeometrySubsets {
                              SceneImportOptions options) {
       var mf = ImporterBase.GetOrAddComponent<MeshFilter>(go);
       var mr = ImporterBase.GetOrAddComponent<MeshRenderer>(go);
-      if (mf.sharedMesh == null) {
-        mf.sharedMesh = new Mesh();
+      if (mf.sharedMesh == null)
+      {
```
jcowles, line 245: We should probably format the entire code base to conform to Unity conventions, in the mean time we should keep it self-consistent (just noticing the brace placement).

package/com.unity.formats.usd/Runtime/Scripts/IO/Geometry/MeshImporter.cs
```diff
@@ -241,8 +241,10 @@ public class GeometrySubsets {
...
+        mf.sharedMesh = new Mesh {name = Guid.NewGuid().ToString()};
+        mf.sharedMesh.MarkDynamic();
```
jcowles, line 247: Ideally, this would only be set if the mesh were animated in the USD file, given how many static meshes a USD set can have.

If not for this change, maybe just add a TODO comment here?

package/com.unity.formats.usd/Runtime/Scripts/IO/Geometry/MeshImporter.cs
clusty: I agree. Call it something maybe like go.name+"_dyn_mesh" or whatever.

judubu: Did that before. But that doesn't help if you want to know if the mesh is shared or not.
I'll probably go for something like: go.name + guid.

package/com.unity.formats.usd/Runtime/Scripts/IO/Geometry/MeshImporter.cs
judubu: Absolutely. I changed my IDE prefs t oavoid that in the future.

package/com.unity.formats.usd/Runtime/Scripts/IO/Geometry/MeshImporter.cs
judubu: Done. Here and in the buildSkinnedMesh.

package/com.unity.formats.usd/Runtime/Scripts/IO/Geometry/MeshImporter.cs
judubu: Done

package/com.unity.formats.usd/Runtime/Scripts/IO/Geometry/MeshImporter.cs
```diff
@@ -222,10 +226,17 @@ public class GeometrySubsets {
...
       var smr = ImporterBase.GetOrAddComponent<SkinnedMeshRenderer>(go);
       if (smr.sharedMesh == null) {
-        smr.sharedMesh = new Mesh();
+        smr.sharedMesh = new Mesh {name = UniqueMeshName(go.name)};
+      }
+            
+      // We only check if a mesh is dynamic when scene.IsPopulatingAccessMask is True. It only happens when a playable is
+      // created, potentially way after mesh creation.
+      if (isDynamic) {
+        smr.sharedMesh.MarkDynamic();
```
jcowles, line 239: Nice!

PR #166, opened 2020-04-23
Release RC-1.0.3-preview.2

judubu:


jcowles:
If you haven't already, we should update the release notes to include the IL2CPP build support

PR #167, opened 2020-04-24
Export textures from packages

hybridherbst:
As we move more and more of our code and data into packages, assets (including textures) are also in packages. This PR fixes USD / USDZ texture export pathes for that case.


jcowles (approved):
Looks good, thanks!

PR #168, opened 2020-04-24
Support texture wrap modes

hybridherbst:
Texture wrap modes were not supported so far. This e.g. fails in the default sample scenes for HDRP / URP, where those are used a lot.

The test data is in the "Tests" folder. Note that exporting textures from packages relies on #167, or you have to copy the data into Assets first.

**Before:**
(note that the default USD texture wrap mode does not even exist as an option in Unity.)
20200424-152040_python

**After:**
20200424-151758_python
20200424-151819_python
20200424-151836_python


package/com.unity.formats.usd/Runtime/Scripts/IO/Materials/ShaderExporterBase.cs
```diff
@@ -60,6 +63,22 @@ public class ShaderExporterBase {
 #endif
     }
 
+    static TextureReaderSample.WrapMode WrapMode(TextureWrapMode wrap)
```
jcowles: This probably belongs in TextureReaderSample, since the conversion is likely to be needed again.

I can move after merge, if you prefer, just let me know.

jcowles (approved):
One function should be moved and it would be nice to get this test to run in CI, but we can do both after merge.

This is awesome, thanks!

package/com.unity.formats.usd/Runtime/Scripts/IO/Materials/ShaderExporterBase.cs
hybridherbst: Done

jcowles:
Merged as #180

PR #169, opened 2020-04-24
Texture value scale

hybridherbst:
USD calls "multiplying a color by a texture" texture scale (a bit confusing imho).
This is currently not supported in usd-unity-sdk, basically breaking a lot of very typical Unity workflows:
image

Note usdview does not support it either (repoted bug https://github.com/PixarAnimationStudios/USD/issues/1129):
image

But AR QuickLook does support it (for color/emissive, also not for metallic/specular):
image


package/com.unity.formats.usd/Runtime/Scripts/IO/Materials/ShaderExporterBase.cs
```diff
@@ -51,6 +52,8 @@ public class ShaderExporterBase {
       uvReader.varname.defaultValue = new TfToken("st");
       scene.Write(usdShaderPath + "/uvReader", uvReader);
       var tex = new TextureReaderSample(filePath, usdShaderPath + "/uvReader.outputs:result");
+      if(scale != Vector4.one)
+        tex.scale = new Connectable<Vector4>(scale);
```
jcowles, line 56: Our coding convention is to always use braces.

jcowles (changes requested):
This is awesome, and thank you for the test cases!

jcowles (approved):

PR #170, opened 2020-04-24
Fix quicklook animation export

hybridherbst:
While usd-unity-sdk properly implements timecode-based USDZ export as far as I can tell, unfortunately the current version of AR QuickLook in iOS doesn't and only supports timecode export.
This PR works around that by exporting USDZ files with a FrameRate of 60 and multiplying all time values by 60. As per USD spec the result is identical - and QuickLook now properly displays the animation as well.

(if you want to see the QuickLook bug in action, there's
a minimal example here https://oval-south-supermarket.glitch.me/ and more bugs we found (and reported to Apple) here: 
https://pfc-ar-quicklook-bugs.glitch.me/)

This comes with example data to see the changed behaviour.

**Note:** there seems to still be an issue with Timeline-based export when the export clip starts right at the beginning of the timeline - frames are then skipped on export. I think this might be "by design" in Unity but is very weird (e.g. Autoplay in Timeline will skip beginning since Unity is not completely done loading by then).
This is visible in the USDA file:
image
but is a separate issue from this PR and should be handled separately.

### DIFF
--- a/package/com.unity.formats.usd/Runtime/Scripts/Timeline/UsdRecorderBehaviour.cs
+++ b/package/com.unity.formats.usd/Runtime/Scripts/Timeline/UsdRecorderBehaviour.cs
@@ -23,6 +23,8 @@ using UnityEditor;
 namespace Unity.Formats.USD {
   public class UsdRecorderBehaviour : PlayableBehaviour {
 
+    // Conversion to keyframes (60 frames per second) to work around QuickLook bug
+    const int kExportFrameRate = 60;
     bool m_isPaused = false;
     public UsdRecorderClip Clip;
 
@@ -67,8 +69,12 @@ namespace Unity.Formats.USD {
         //  m_usdScene.Stage.SetTimeCodesPerSecond(1);
         //}
 
-        Clip.UsdScene.FrameRate = 60;
-        Clip.UsdScene.Stage.SetFramesPerSecond(60);
+        // Regardless of the actual sampling rate (e.g. Timeline playback speed), we are converting
+        // the timecode from seconds to frames with a sampling rate of 60 FPS. This has the nice quality
+        // of adding additional numerical stability.
+        // In the event that the timeline is not configured for 60 FPS playback, we rely on USD's linear
+        // interpolation mode to up-sample to 60 FPS.
+        Clip.UsdScene.FrameRate = kExportFrameRate ;
         Clip.UsdScene.Stage.SetInterpolationType(pxr.UsdInterpolationType.UsdInterpolationTypeLinear);
 
         // For simplicity in this example, adding game objects while recording is not supported.
@@ -78,7 +84,7 @@ namespace Unity.Formats.USD {
         Clip.Context.activePolicy = Clip.m_activePolicy;
         Clip.Context.exportMaterials = Clip.m_exportMaterials;
 
-        Clip.UsdScene.StartTime = currentTime * 60;
+        Clip.UsdScene.StartTime = currentTime * kExportFrameRate ;
 
         // Export the "default" frame, that is, all data which doesn't vary over time.
         Clip.UsdScene.Time = null;
@@ -104,7 +110,7 @@ namespace Unity.Formats.USD {
       }
 
       Clip.Context = new ExportContext();
-      Clip.UsdScene.EndTime = currentTime * 60;
+      Clip.UsdScene.EndTime = currentTime * kExportFrameRate ;
 
       // In a real exporter, additional error handling should be added here.
       if (!string.IsNullOrEmpty(Clip.m_usdFile)) {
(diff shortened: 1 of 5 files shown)


package/com.unity.formats.usd/Runtime/Scripts/Timeline/UsdRecorderBehaviour.cs
```diff
@@ -67,9 +67,9 @@ public class UsdRecorderBehaviour : PlayableBehaviour {
         //  m_usdScene.Stage.SetTimeCodesPerSecond(1);
         //}
 
-        // TODO: How does one extract the time mode (frames or seconds) from the Timeline?
-        Clip.UsdScene.Stage.SetFramesPerSecond(30);
-        Clip.UsdScene.Stage.SetTimeCodesPerSecond(1);
+        Clip.UsdScene.FrameRate = 60;
+        Clip.UsdScene.Stage.SetFramesPerSecond(60);
```
jcowles: Why do you also call SetFramesPerSecond here?
Assigning to FrameRate will set both TimeCodesPerSecond as well as FramesPerSecond

package/com.unity.formats.usd/Runtime/Scripts/Timeline/UsdRecorderBehaviour.cs
```diff
@@ -78,7 +78,7 @@ public class UsdRecorderBehaviour : PlayableBehaviour {
         Clip.Context.activePolicy = Clip.m_activePolicy;
         Clip.Context.exportMaterials = Clip.m_exportMaterials;
 
-        Clip.UsdScene.StartTime = currentTime;
+        Clip.UsdScene.StartTime = currentTime * 60;
```
jcowles: We should add a constant for the frame rate (e.g. rather than copying "60" around)

e.g. declare
`const int kExportFrameRate = 60;`

At the top of this file

package/com.unity.formats.usd/Runtime/Scripts/Timeline/UsdRecorderBehaviour.cs
```diff
@@ -129,7 +129,7 @@ public class UsdRecorderBehaviour : PlayableBehaviour {
         Debug.LogError("Process: context.scene is null");
       }
 
-      Clip.UsdScene.Time = currentTime;
+      Clip.UsdScene.Time = currentTime * 60; // conversion to keyframes to work around QuickLook bug
```
jcowles: If you want to add this comment, I would put it near the kExportFrameRate constant.

Also, nit pick: we typically don't put comments on the same line as the code.

package/com.unity.formats.usd/Runtime/Scripts/Timeline/UsdRecorderBehaviour.cs
```diff
@@ -67,9 +67,9 @@ public class UsdRecorderBehaviour : PlayableBehaviour {
         //  m_usdScene.Stage.SetTimeCodesPerSecond(1);
         //}
 
-        // TODO: How does one extract the time mode (frames or seconds) from the Timeline?
-        Clip.UsdScene.Stage.SetFramesPerSecond(30);
-        Clip.UsdScene.Stage.SetTimeCodesPerSecond(1);
+        Clip.UsdScene.FrameRate = 60;
+        Clip.UsdScene.Stage.SetFramesPerSecond(60);
+        Clip.UsdScene.Stage.SetInterpolationType(pxr.UsdInterpolationType.UsdInterpolationTypeLinear);
```
jcowles, line 78: Please add a comment here and note that:

```
Regardless of the actual sampling rate (e.g. Timeline playback speed), we are converting
the timecode from seconds to frames with a sampling rate of 60 FPS. This has the nice quality
of adding additional numerical stability.

In the event that the timeline is not configured for 60 FPS playback, we rely on USD's linear
interpolation mode to up-sample to 60 FPS.
```

jcowles (changes requested):
This needs a few minor changes, but I think overall this is a great improvement, thanks!

package/com.unity.formats.usd/Runtime/Scripts/Timeline/UsdRecorderBehaviour.cs
hybridherbst: Added.

Note that I'm not 100% sure that this actually adds numerical stability; the timeline is just too "off beat" to produce exact, int-aligned numbers here. You can see some exported timings in the screenshot in my PR. For exact int-aligned frame numbers (and preventing that initial gap) more work would be needed.

jcowles: Ah, I see, well in theory it should enable us to put key frames at exact integer frame offsets (e.g. for offline export), but yeah, it's probably not going to help without putting some effort on making that more consistent.

jcowles (approved):

jcowles:
I believe this also fixes #82

hybridherbst:
Still have to test. But I definitely did a successful runtime export in the past, just not sure in which repo state. Also will have to test for #100 separately.

PR #171, opened 2020-04-24
USDZ timeline export in UsdRecorderClip

hybridherbst:
A common functionality for us for usd-unity-sdk is exporting content for use in Apple AR QuickLook.
For this, the menu item ("export as USDZ") is not sufficient as it does not export animation.
Additionally, exporting the timeline as USD and converting with the usdtools is also cumbersome, as it requires to remember to scale the object by 100x in Unity (since usdztools don't do the scale conversion).

To remedy this, I added the ability to UsdRecorderClip to also export USDZ (based on the file extension currently). This uses the same principles of changing working directory, packing in the end, and scaling the object by 100x. Resulting USDZ files are ready to be used directly in AR QuickLook in correct and expected scale (1 unity unit = 1 meter).

This could definitely be nicer in terms of UX (e.g. a file type dropdown for the clip), but it does the job and is (for us) a considerable QoL improvement.

Quite some code changes for the effect of having a single different letter :)
image

### DIFF
--- a/package/com.unity.formats.usd/Runtime/Scripts/Timeline/UsdRecorderBehaviour.cs
+++ b/package/com.unity.formats.usd/Runtime/Scripts/Timeline/UsdRecorderBehaviour.cs
@@ -15,6 +15,8 @@
 using UnityEngine;
 using UnityEngine.Playables;
 using USD.NET;
+using System.IO;
+using pxr;
 
 #if UNITY_EDITOR
 using UnityEditor;
@@ -25,6 +27,12 @@ namespace Unity.Formats.USD {
 
     bool m_isPaused = false;
     public UsdRecorderClip Clip;
+    string usdcFileName;
+    string usdzFileName;
+    string usdzFilePath;
+    string currentDir;
+    DirectoryInfo tmpDir;
+    GameObject _root;
 
     // ------------------------------------------------------------------------------------------ //
     // Recording Control.
@@ -32,6 +40,7 @@ namespace Unity.Formats.USD {
 
     public void BeginRecording(double currentTime, GameObject root) {
       InitUsd.Initialize();
+      _root = root;
 
       if (!root) {
         Debug.LogError("ExportRoot not assigned.");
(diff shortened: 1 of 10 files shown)


package/com.unity.formats.usd/Runtime/Scripts/Timeline/UsdRecorderBehaviour.cs
```diff
@@ -82,6 +109,11 @@ public class UsdRecorderBehaviour : PlayableBehaviour {
 
         // Export the "default" frame, that is, all data which doesn't vary over time.
         Clip.UsdScene.Time = null;
+
+        // USDZ is in centimeters.
+        if (Clip.IsUSDZ)
+          root.transform.localScale = localScale * 100;
```
jcowles: Nit: add braces

More important: can we make the scale part of the ExportContext and handle it here?
https://github.com/Unity-Technologies/usd-unity-sdk/blob/fde5076004a8d4dc1e852adb3ffa334872dbaed1/package/com.unity.formats.usd/Runtime/Scripts/IO/Scene/SceneExporter.cs#L149

So that it's symetrical with the way we handle it for SceneImport:
https://github.com/Unity-Technologies/usd-unity-sdk/blob/fde5076004a8d4dc1e852adb3ffa334872dbaed1/package/com.unity.formats.usd/Runtime/Scripts/IO/Scene/SceneImportOptions.cs#L115

package/com.unity.formats.usd/Runtime/Scripts/Timeline/UsdRecorderBehaviour.cs
```diff
@@ -93,7 +125,14 @@ public class UsdRecorderBehaviour : PlayableBehaviour {
           Clip.UsdScene.Close();
           Clip.UsdScene = null;
         }
+        if (Clip.IsUSDZ)
+          usdzTemporaryDir.Delete(recursive: true);
         throw;
+      } finally {
+        // USDZ is in centimeters.
+        if (Clip.IsUSDZ)
+          root.transform.localScale = localScale;
```
jcowles: This goes away if handled in the exporter

package/com.unity.formats.usd/Runtime/Scripts/Timeline/UsdRecorderBehaviour.cs
```diff
@@ -131,7 +193,12 @@ public class UsdRecorderBehaviour : PlayableBehaviour {
 
       Clip.UsdScene.Time = currentTime;
       Clip.Context.exportMaterials = false;
+      var localScale = root.transform.localScale;
+      if (Clip.IsUSDZ)
+        root.transform.localScale = localScale * 100;
```
jcowles: This goes away if handled in the exporter

package/com.unity.formats.usd/Runtime/Scripts/Timeline/UsdRecorderBehaviour.cs
```diff
@@ -131,7 +193,12 @@ public class UsdRecorderBehaviour : PlayableBehaviour {
...
       SceneExporter.Export(root, Clip.Context, zeroRootTransform: false);
+      if (Clip.IsUSDZ)
+        root.transform.localScale = localScale;
```
jcowles: This goes away if handled in the exporter

jcowles (changes requested):
Nice work!
This is an awesome contribution :D

package/com.unity.formats.usd/Runtime/Scripts/Timeline/UsdRecorderBehaviour.cs
hybridherbst: That's definitely a better place.

jcowles (approved):

jcowles:
Merged as #181

PR #172, opened 2020-04-24
Export dynamic textures

hybridherbst:
Currently, textures are only exported from disk. This creates issues with a number of typical Unity workflows, especially for USDZ export where only PNG and JPEG are supported:

1. PSD files or other unsupported file formats
2.  .asset files (texture data generated from script in Unity)
3. RenderTextures that are filled at runtime
4. Normals generated from grayscale textures (the original files will be copied)

This PR fixes 1-3. A fix for 4 will follow in a later PR.

Basically, unsupported textures will be converted to PNG by blitting them into an ARGB32 texture and getting that data back to the CPU. This also avoids the need to have textures marked as readable, and also allows texture exporting at runtime.


package/com.unity.formats.usd/Runtime/Scripts/IO/Materials/ShaderExporterBase.cs
```diff
@@ -33,33 +33,120 @@ public class ShaderExporterBase {
...
+        fileName = System.IO.Path.GetFileName(srcPath);
+        filePath = System.IO.Path.Combine(destTexturePath, fileName);
+
+        if(System.IO.File.Exists(srcPath))
+        { 
+          // USD officially only supports png / jpg / jpeg
+          // https://graphics.pixar.com/usd/docs/Usdz-File-Format-Specification.html
 
+          var ext = System.IO.Path.GetExtension(srcPath).ToLowerInvariant();
+          if(ext == ".png" || ext == ".jpg" || ext == ".jpeg") {
```
jcowles, line 75: I can't find the code now, but I feel like USD supports the esoteric jpeg extensions, like jfif, etc.

package/com.unity.formats.usd/Runtime/Scripts/IO/Materials/ShaderExporterBase.cs
```diff
@@ -33,33 +33,120 @@ public class ShaderExporterBase {
...
+          fileName = texture.name + "_png";
+  #endif
+        filePath = System.IO.Path.Combine(destTexturePath, fileName + ".png");
+        
+        // TODO extra care has to be taken of Normal Maps etc., since these are in a converted format in memory (namely, 16 bit G and A instead of 8 bit RGBA)
+        // An example of this conversion in a shader is in Khronos' UnityGLTF implementation.
+        // basically, the blit has do be done with the right unlit conversion shader to get a proper "file-based" tangent space normal map back
+
+        // Blit the texture and get it back to CPU
+        // Note: Can't use RenderTexture.GetTemporary because that doesn't properly clear alpha channel
```
jcowles, line 99: It's probably a bad idea to rely on any RenderTexture to be initialized to zeros, especially if the pipeline is configured to be a modern explicit graphics API.

package/com.unity.formats.usd/Runtime/Scripts/IO/Materials/ShaderExporterBase.cs
```diff
@@ -33,33 +33,120 @@ public class ShaderExporterBase {
...
+        fileName = texture.name + "_" + Random.Range(10000000, 99999999).ToString();
+  #if UNITY_EDITOR
+        if (UnityEditor.AssetDatabase.Contains(texture))
+          fileName = texture.name + "_png";
+  #endif
+        filePath = System.IO.Path.Combine(destTexturePath, fileName + ".png");
+        
+        // TODO extra care has to be taken of Normal Maps etc., since these are in a converted format in memory (namely, 16 bit G and A instead of 8 bit RGBA)
+        // An example of this conversion in a shader is in Khronos' UnityGLTF implementation.
+        // basically, the blit has do be done with the right unlit conversion shader to get a proper "file-based" tangent space normal map back
```
jcowles: Might also be worth noting that Unity currently only supports 8-bit texture exports (16-bit format support is in the works).

package/com.unity.formats.usd/Runtime/Scripts/IO/Materials/ShaderExporterBase.cs
```diff
@@ -33,33 +33,120 @@ public class ShaderExporterBase {
...
+          RenderTexture.active = rt;
+          resultTex.ReadPixels(new Rect(0, 0, texture.width, texture.height), 0, 0);
+          resultTex.Apply();
+
+          System.IO.File.WriteAllBytes(filePath, resultTex.EncodeToPNG());
+          if (System.IO.File.Exists(filePath))
+            textureIsExported = true;
+        }
+        finally {
+          RenderTexture.active = null;
```
jcowles: I learned recently that you are expected to restore the original value of `RenderTexture.active`, setting it to `null` will be incorrect in some contexts (though /probably/ safe, there are corner cases where it is not).

package/com.unity.formats.usd/Runtime/Scripts/IO/Materials/ShaderExporterBase.cs
```diff
@@ -33,33 +33,120 @@ public class ShaderExporterBase {
...
+
+      if(!textureIsExported)
+      {
+        var tex = new Texture2D(1, 1, TextureFormat.ARGB32, true);
+        try { 
+          tex.SetPixel(0, 0, Color.white);
+          tex.Apply();
+          System.IO.File.WriteAllBytes(filePath, tex.EncodeToPNG());
+          if (System.IO.File.Exists(filePath))
+              textureIsExported = true;
```
jcowles: Nit: please add braces

package/com.unity.formats.usd/Runtime/Scripts/IO/Materials/ShaderExporterBase.cs
```diff
@@ -33,33 +33,120 @@ public class ShaderExporterBase {
...
+        }
+        finally
```
jcowles: Nit: braces go on the same line

package/com.unity.formats.usd/Runtime/Scripts/IO/Materials/ShaderExporterBase.cs
```diff
@@ -33,33 +33,120 @@ public class ShaderExporterBase {
...
+
+        var uvReader = new PrimvarReaderSample<Vector2>();
+        uvReader.varname.defaultValue = new TfToken("st");
+        scene.Write(usdShaderPath + "/uvReader", uvReader);
+        var tex = new TextureReaderSample(filePath, usdShaderPath + "/uvReader.outputs:result");
+        scene.Write(usdShaderPath + "/" + textureName, tex);
+        return usdShaderPath + "/" + textureName + ".outputs:" + textureOutput;
+      }
+      else
+        throw new System.Exception("Texture wasn't exported.");
```
jcowles: Nit: add braces.

Question: Is this truly an exception or do we expect it to happen sometimes?

jcowles (changes requested):
Another awesome improvement -- thank you so much for taking the time to submit these PRs!!

package/com.unity.formats.usd/Runtime/Scripts/IO/Materials/ShaderExporterBase.cs
hybridherbst: True, USD also supports TGA/TIF/PSD etc.
But USDZ only supports PNG/JPG (https://graphics.pixar.com/usd/docs/Usdz-File-Format-Specification.html), so best way would be to distinguish here in which kind of export we currently are (ExportContext?) - what would be the best way to approach that?

package/com.unity.formats.usd/Runtime/Scripts/IO/Materials/ShaderExporterBase.cs
hybridherbst: added a GL.Clear

package/com.unity.formats.usd/Runtime/Scripts/IO/Materials/ShaderExporterBase.cs
hybridherbst: For the future work outlined here it's more a matter that the channels are shifted (Normal texture is stored differently depending on platform) and that this needs to be taken into account when converting back to an "expected" tangent space representation. Refined the comment a bit but still think this should be tackled in a later PR / filed as known issue.

package/com.unity.formats.usd/Runtime/Scripts/IO/Materials/ShaderExporterBase.cs
hybridherbst: This was just copied over from the previous implementation - I changed it to an error log and added some more sane help to the user what actually went wrong.

hybridherbst: It was happening all the time before my changes, now it should happen much less often:
- if the texture is null
- if it's a 3D texture that can't be blitted
- something's wrong on disk write (e.g. a known issue of mine is that if the export path is exactly where the src path is this will obviously fail but could be handled better - not covered by this PR).

package/com.unity.formats.usd/Runtime/Scripts/IO/Materials/ShaderExporterBase.cs
jcowles: Oh, I just meant that "jfif" IS actually a jpeg file, it's just an unusual extension that most people don't know about.

I think it's actually fine to only export png/jpg because I've seen issues with USD loading files like PDFs anyway -- sometimes it works, sometimes not (maybe bugs in OpenImageIO).

Anyway, I think this is a good default without adding more options to ExportContext, I'm just suggesting to add all the supported JPEG file extensions.

package/com.unity.formats.usd/Runtime/Scripts/IO/Materials/ShaderExporterBase.cs
```diff
@@ -33,33 +33,123 @@ public class ShaderExporterBase {
...
+          }
+        }
+      }
+  #endif
+      if (!textureIsExported)
+      {
+        fileName = texture.name + "_" + Random.Range(10000000, 99999999).ToString();
+  #if UNITY_EDITOR
+        if (UnityEditor.AssetDatabase.Contains(texture))
+          fileName = texture.name + "_png";
```
jcowles: Oh, I just noticed -- need braces here

Also, maybe add a comment to explain that this is just to avoid file collisions in the event that there already exists a file to export by the same name.

package/com.unity.formats.usd/Runtime/Scripts/IO/Materials/ShaderExporterBase.cs
hybridherbst: I see. In above spec link it's mentioned: "png and jpeg files (any of the multiple common extensions for jpeg)".
I still think this needs to be configurable (maybe not in this PR) since otherwise some roundtrip stuff breaks (moving usd out of Unity and back in will have different textures), not sure how important that is.

package/com.unity.formats.usd/Runtime/Scripts/IO/Materials/ShaderExporterBase.cs
```diff
@@ -33,33 +33,123 @@ public class ShaderExporterBase {
...
+        try { 
+          RenderTexture.active = rt;
+          GL.Clear(true, true, Color.clear);
+          Graphics.Blit(texture, rt);
+          resultTex.ReadPixels(new Rect(0, 0, texture.width, texture.height), 0, 0);
+          resultTex.Apply();
+
+          System.IO.File.WriteAllBytes(filePath, resultTex.EncodeToPNG());
+          if (System.IO.File.Exists(filePath))
+            textureIsExported = true;
```
judubu: Missing braces here.

judubu (changes requested):
Thanks again for all those PRs, this is great.

Beside the missing braces, this looks good to me.

package/com.unity.formats.usd/Runtime/Scripts/IO/Materials/ShaderExporterBase.cs
hybridherbst: Added the comment. Also, I removed the #if UNITY_EDITOR part because now that I think about it this will cause more trouble than help - textures with identical names in the AssetDB would override / collide, so I removed it again for now. Ideas for future improvement are in the TODO section in the comment.

jcowles:
Merged as #182

PR #183, opened 2020-05-14
Fix animation export end time regression

hybridherbst:
Missing multiplication by kExportFrameRate resulting in truncated exports.

### DIFF
--- a/package/com.unity.formats.usd/Runtime/Scripts/Timeline/UsdRecorderBehaviour.cs
+++ b/package/com.unity.formats.usd/Runtime/Scripts/Timeline/UsdRecorderBehaviour.cs
@@ -145,14 +145,14 @@ namespace Unity.Formats.USD {
       }
 
       Clip.Context = new ExportContext();
-      Clip.UsdScene.EndTime = currentTime * kExportFrameRate ;
+      Clip.UsdScene.EndTime = currentTime * kExportFrameRate;
 
       try {
         if(Clip.IsUSDZ && usdzTemporaryDir != null)
           Directory.SetCurrentDirectory(usdzTemporaryDir.FullName);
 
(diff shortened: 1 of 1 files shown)


package/com.unity.formats.usd/Runtime/Scripts/Timeline/UsdRecorderBehaviour.cs
```diff
@@ -145,14 +145,14 @@ public class UsdRecorderBehaviour : PlayableBehaviour {
...
-      Clip.UsdScene.EndTime = currentTime * kExportFrameRate ;
+      Clip.UsdScene.EndTime = currentTime * kExportFrameRate;
 
       try {
         if(Clip.IsUSDZ && usdzTemporaryDir != null)
           Directory.SetCurrentDirectory(usdzTemporaryDir.FullName);
 
         Clip.Context = new ExportContext();
-        Clip.UsdScene.EndTime = currentTime;
+        Clip.UsdScene.EndTime = currentTime * kExportFrameRate;
```
judubu, line 152: Looks like something was messed up during the PR merge and the ExportContext creation is now done twice. Can you remove the l147-148 please.

hybridherbst: Done. Turns out the kExportFrameRate error was also a merge issue probably, as that was also double (removed it as well).

For my understanding, why is the Clip.Context re-created in StopRecording at all?

jcowles: No, seems suspicious to me...

jcowles (approved):

PR #184, opened 2020-05-14
Fix ExportSkeleton USDZ scale

hybridherbst:
After adding scale in ExportContext bounds were now broken for animated meshes. I'm still not 100% sure about this, but at least the order of magnitude of the bounds is correct again (and exports look right in usdview which they don't without this PR).

### DIFF
--- a/package/com.unity.formats.usd/Runtime/Scripts/IO/Skel/SkeletonExporter.cs
+++ b/package/com.unity.formats.usd/Runtime/Scripts/IO/Skel/SkeletonExporter.cs
@@ -80,6 +80,8 @@ namespace Unity.Formats.USD {
       // This is required because USD expects the extent to be a local bound for the SkelRoot.
       var oldParent = objContext.gameObject.transform.parent;
       objContext.gameObject.transform.SetParent(null, worldPositionStays: false);
+      var originalScale = objContext.gameObject.transform.localScale;
+      objContext.gameObject.transform.localScale *= 1 / exportContext.scale;
 
       try {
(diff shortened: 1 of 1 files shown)


package/com.unity.formats.usd/Runtime/Scripts/IO/Skel/SkeletonExporter.cs
```diff
@@ -80,6 +80,8 @@ public static class SkeletonExporter {
       // This is required because USD expects the extent to be a local bound for the SkelRoot.
       var oldParent = objContext.gameObject.transform.parent;
       objContext.gameObject.transform.SetParent(null, worldPositionStays: false);
+      var originalScale = objContext.gameObject.transform.localScale;
+      objContext.gameObject.transform.localScale *= 1 / exportContext.scale;
```
jcowles, line 84: I do believe this was intended to be captured by `SetParent(null, ...)` above, it's just that the scale wasn't captured.

jcowles (approved):
Looks good, thanks for the fix!

package/com.unity.formats.usd/Runtime/Scripts/IO/Skel/SkeletonExporter.cs
jcowles: I guess my remaining question is if this should be `1/exportContext.scale` or if it should just be `1`.

jcowles (changes requested):
After some local testing, this only seems right in some scenarios... I'm digging in a bit more to better understand what's going on between the bbox and geometry scales

PR #194, opened 2020-07-04
Add PBR support to standard shader

### LINKED ISSUE #191: Exported Usdz does not have metallic and roughness map.
Hey guys, 
I am new in working with usdz exporter. As I always do with Unity, I am using grayscale Metallic map with roughness in the alpha channel. 

Screenshot 2020-06-11 at 18 44 10
Screenshot 2020-06-11 at 18 44 23

But for some reason, the exported usdz does not have the metallic/roughness info. It is just flat 50% smoothness and 0 metallic (default settings, I believe).
Screenshot 2020-06-11 at 18 45 52

Is there a specific way I need to add these maps in Unity to get it into usdz?
Any help is welcome.

hybridherbst:
Fixes #191 

@jcowles , I extended the ChannelCombiner shader to allow for arbitrary export configurations, and for now went with
(MetallicSmoothness) R, _, _, A => _, 1-A ,R, 1

This might be desired, but based on pipeline this could also be changed to keep the R+A format and just invert alpha. I went with this because it allows jpg compression of the resulting texture which is important for keeping USDZ file size small.

Also @jcowles I did my best to adapt the HDRPShaderImporter (the only place where ChannelCombiner was used) to the updated shader, but didn't really test if HDRP export now still works as expected, would be great if you could verify that (or are there tests for it?).

Additionally, I added two example objects; the well-known DamagedHelmet that I converted to Unity, and a multi material sample that I created.
image


jcowles:
This is super awesome! Will take a look ASAP and test in HDRP / UniversalRP.

jcowles:
Yeah, not looking quite right when exported from HDRP (left is HDRP render, right is usdview):
image


Will dig in to see what's going on

jcowles:
Hmm, albedo looks better when exporting from built-in:
image


However the roughness doesn't look correct -- note that the specular highlight in usdview should be a very rough area of the the sphere:
image

jcowles:
I think this is actually a bug in the USD texture loader, or something. Following up on usd-interest.

Thread started here:
https://groups.google.com/forum/#!topic/usd-interest/gr_xmas5zFA

hybridherbst:
Uuuugh please don't use usdview for verifying correctness of USD files :(

There's a TON of open bugs, UsdPreviewSurface isn't actually officially supported in usdview with any of the built-in renderer backends!

- PBR doesn't fully work (as you just noticed)
- Transparency doesn't fully work (https://github.com/PixarAnimationStudios/USD/issues/990)
- Most attributes don't work (e.g. scale/bias are incorrect https://github.com/PixarAnimationStudios/USD/issues/1129)
- Vertex color doesn't always work
- Interpolation types are not properly supported
- etc etc.

I usually verify correctness either in an external tool such as Houdini or directly in RealityComposer on Mac / QuickLook on iOS.

drwave:
> On Jul 4, 2020, at 1:45 PM, hybridherbst <notifications@github.com> wrote:
> 
> 
> Uuuugh please don't use usdview for verifying correctness of USD files :(
> 
> There's a TON of open bugs, UsdPreviewSurface isn't actually officially supported in usdview with any of the built-in renderer backends!
> 
> PBR doesn't fully work (as you just noticed)
> Transparency doesn't fully work (PixarAnimationStudios/USD#990 <https://github.com/PixarAnimationStudios/USD/issues/990>)
> Most attributes don't work (e.g. scale/bias are incorrect PixarAnimationStudios/USD#1129 <https://github.com/PixarAnimationStudios/USD/issues/1129>)
> Vertex color doesn't always work
> Interpolation types are not properly supported
> etc etc.
> I usually verify correctness either in an external tool such as Houdini or directly in RealityComposer on Mac / QuickLook on iOS.
> 
Agreed.

In my experience, HoudiniGL & Karma in Houdini 18 are the most accurate shipping implementations of UsdPreviewSurface, with ARQuickLook in the Apple ecosystem good as well.

Storm is getting better, but it’s not as accurate as those others.

I've actually been most disappointed with RenderMan’s support for UsdPreviewSurface - assets that look great in Houdini or ARQuickLook, and look understandable in Storm are a disaster in RenderMan. I logged some bugs with RenderMan support, but never got any satisfactory answers. 

Hoping to loop back in on this once 20.08 gets released and see where Storm is at.

jcowles:
Agreed, however I understand the limitations of usdview/storm pretty well and I have a secondary goal of calling out limitations and bugs I find there, so it's always my personal go-to. I also don't expect it to look perfect.

I didn't realize Houdini had gotten so good, I'll start using that as well.

The good news is, it looks like this bug was fixed in 20.05 and usdview also (finally) has an environment light :)

Now that I have a 20.05 build up and running locally, I'll continue debugging what's happening with the HDRP export, which still looks epicly broken in usdview.

jcowles:
Seems to be several problems with the HDRP export:
 * Emissive is not bound to any texture and set to full (1,1,1), which is why the helmet is solid white
 * Roughness seems inverted
 * Opacity - looks like a storm issue
 * Displacement is connected and has non-zero values in the blue channel, so maybe this is expected? However the displacement is causing cracks, which is probably not expected

hybridherbst:
Opacity was an actual export issue. Both usdview and ARQL will switch to alpha modes when opacity is connected, and to cutoff/mask when opacityThreshold is connected. I bound these to the proper shader keywords now, and opacity is working, so now opacity will not be connected at all when not used by the material.

Also, I guarded displacement export against the right shader keywords so it's only exported when displacement modes are actually used in the shader. Note that I did a simple version (no texture bias/scale) - fully getting this "right" would be way more complex; Unity does a lot of magic with the _HeightMap values for HDRP (there's 9 different shader floats controlling the behaviour depending on MinMax or Amplitude mode or remapping or ...). For a first pass it should suffice (aka displacement is exported if the shader uses displacement or tesselation).

Roughness seems to work correctly but overall everything is way too dark, might be a linear/gamma thing?

Where we're at vs. where we want to be (in ARQL):
Image from iOS
Image from iOS (1)

hybridherbst:
Backlog: the rarely used "Smoothness source is Albedo Alpha" option doesn't work yet.

image

jcowles (approved):

PR #196, opened 2020-07-16
Fix subdivision scheme and normal interpolation on export

### LINKED ISSUE #192: Exported USD and USDZ have corrupted normals along UV seams (FTV-1110)
I've been trying to find the root cause for this myself but haven't been successful in doing so.
UV seams in meshes cause visible seams along normals in USD and USDZ files exported with usd-unity.

It's clear that vertices need to be split in places where UV seams are but the resulting normals should be identical, resulting in smooth shading independent of UV seams.

My assumption so far is that it has something to do with normal interpolation behaviour; Blender sets that explicitly to "faceVarying" while Unity leaves it as default.

Reproduction cases:
https://glitch.com/edit/#!/usd-normals-bug
(you can directly download the usdz files from the "assets" section in Glitch)

Reproduction from Unity:
[NormalsBugs.unitypackage.zip](https://github.com/Unity-Technologies/usd-unity-sdk/files/4872322/NormalsBugs.unitypackage.zip)

Import these, export as USD or USDZ.
Import them in Blender or anything else that can export USDZ and try again.

Note that re-importing the corrupt files in Unity looks correct, so some logic seems to be wrong in both import and export.

Ground truth in Unity (these are smoothed cubes; the left one has 3x3 quads per face, the others 1x1. The left one has a deliberately shitty UV layout to show the issues better. UV layout should not affect shading at all if no textures are used.)
image

Exported from Unity:
IMG_0149

Same file, exported from Blender:
IMG_0153

For simple shapes, it's the same, visible edges along all UV seams:
image

For complex shapes, also, but much less visible since there's more going on - this is where I originally found the issue - note the vertial line along the center:
IMG_0154

**Reimporting them with usd-unity looks correct,** which I believe points to different defaults being in the spec and used by Unity for some interpolation:
image

hybridherbst:
Set subdivision scheme to "none" and explicity set normal interpolation to "vertex". This fixes #192 but will need additional changes for "USD roundtrip" scenarios where keeping the subdiv scheme is desired. I think that is a problem for later since such a round trip would also require proper support for faceVarying attributes being carried through / reconstructed on export if desired. 

Current state of this should potentially be added to the docs to make it more clear what "roundtrip" means and what export currently does.

Image from iOS
Image from iOS (1)
Image from iOS

### DIFF
--- a/package/com.unity.formats.usd/Runtime/Scripts/IO/Geometry/MeshExporter.cs
+++ b/package/com.unity.formats.usd/Runtime/Scripts/IO/Geometry/MeshExporter.cs
@@ -295,15 +295,26 @@ namespace Unity.Formats.USD {
         scene.Write(path, sample);
         UnityEngine.Profiling.Profiler.EndSample();
 
+        pxr.UsdPrim usdPrim = scene.GetPrimAtPath(path);
+
         // TODO: this is a bit of a half-measure, we need real support for primvar interpolation.
         // Set interpolation based on color count.
         if (sample.colors != null && sample.colors.Length == 1) {
-          pxr.UsdPrim usdPrim = scene.GetPrimAtPath(path);
           var colorPrimvar = new pxr.UsdGeomPrimvar(usdPrim.GetAttribute(pxr.UsdGeomTokens.primvarsDisplayColor));
           colorPrimvar.SetInterpolation(pxr.UsdGeomTokens.constant);
           var opacityPrimvar = new pxr.UsdGeomPrimvar(usdPrim.GetAttribute(pxr.UsdGeomTokens.primvarsDisplayOpacity));
           opacityPrimvar.SetInterpolation(pxr.UsdGeomTokens.constant);
         }
+
+        // for polygonal meshes, we do not want the default subdivision scheme of "catmull-clark" but instead "none"
+        // so that authored normals keep working.
+        var usdGeomMesh = new pxr.UsdGeomMesh(usdPrim);
+        usdGeomMesh.CreateSubdivisionSchemeAttr().Set(new pxr.TfToken("none"));
+
+        // explicitly set the normal interpolation here –
+        // default should be "vertex" anyways but seems some viewers don't respect the default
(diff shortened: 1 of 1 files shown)


package/com.unity.formats.usd/Runtime/Scripts/IO/Geometry/MeshExporter.cs
```diff
@@ -295,16 +295,27 @@ public static class MeshExporter {
...
           var colorPrimvar = new pxr.UsdGeomPrimvar(usdPrim.GetAttribute(pxr.UsdGeomTokens.primvarsDisplayColor));
           colorPrimvar.SetInterpolation(pxr.UsdGeomTokens.constant);
           var opacityPrimvar = new pxr.UsdGeomPrimvar(usdPrim.GetAttribute(pxr.UsdGeomTokens.primvarsDisplayOpacity));
           opacityPrimvar.SetInterpolation(pxr.UsdGeomTokens.constant);
         }
 
+        // for polygonal meshes, we do not want the default subdivision scheme of "catmull-clark" but instead "none"
+        // so that authored normals keep working.
+        var usdGeomMesh = new pxr.UsdGeomMesh(usdPrim);
+        usdGeomMesh.CreateSubdivisionSchemeAttr().Set(new pxr.TfToken("none"));
```
jcowles, line 312: I think we should probably make this part of the schema class, in USD.NET.Unity.

package/com.unity.formats.usd/Runtime/Scripts/IO/Geometry/MeshExporter.cs
```diff
@@ -295,16 +295,27 @@ public static class MeshExporter {
...
+
+        // explicitly set the normal interpolation here –
+        // default should be "vertex" anyways but seems some viewers don't respect the default
```
jcowles, line 315: Can you note the problematic viewers and versions here, for future reference?

jcowles (changes requested):
Thanks! 

I think we should move subdiv scheme down to the base schema class, let me know if you want to do this or if you would prefer me to do it.

phocodes (approved):

PR #197, opened 2020-07-20
Update USD to 20.08

jcowles:


jcowles:
The Yamato error is also happening with older binaries, so I have to assume it's something funky with trunk (20.2).

I'm looking into it.

judubu:
Thanks for the big update. 
Do you have the resources to build for OSX and Linux? I wanted to get the build infra back up and running mid when I'm back in August.
The trunk jobs on Yamato started failing beginning of June with all errors stating USD.NET not being loadable. Looked into it a little bit but came back empty handed. Completely forgot about it with the Chaplin rush.

jcowles:
I can produce an OSX build (manually via XCode), but I have no local ability to create a Linux build.

I started setting up locally to start to learn our new build infrastructure (for automated win/linux/osx builds), though I'm now waiting for that team to process me as a new user.

PR #213, opened 2020-10-27
USDU 91 maintain material names on import

### LINKED ISSUE #174: GameObject import does not import material names (FTV-661)
While trying to replicate @jcowles import of the Attic USD data set, I found that all imported materials are named "Standard (Clone)" despite the materials having names in the USD file.
Would be great if the name survived import!

vkovec:
## Purpose of this PR

**Ticket/Jira #:** USDU 91

Set the material name in Unity to be the same as the material name in the USD file on import.
Previously all materials would have names like "Standard (Clone)".

image

Material name is taken from the full material path.

## Testing

Added a new automation test and usd asset.
Manual testing:
1. Import a USD asset containing materials (USD > Import as GameObjects)
2. In the USD asset import settings, select "Import Preview Surface" for the Materials setting
3. Refresh

There should be no effect on performance.

## Overall Product Risks

**Complexity:**

Minimal

**Halo Effect:**

Minimal

## Additional information

**Note to reviewers:**

**Reminder:**

### DIFF
--- a/package/com.unity.formats.usd/Runtime/Scripts/IO/Materials/MaterialImporter.cs
+++ b/package/com.unity.formats.usd/Runtime/Scripts/IO/Materials/MaterialImporter.cs
@@ -155,7 +155,11 @@ namespace Unity.Formats.USD {
         matAdapter.ImportFromUsd();
       }
 
-      mat.name = materialPath;
+      // Get the material name from the path
+      if (mat != null && !string.IsNullOrEmpty(materialPath))
+      {
+         mat.name = new pxr.SdfPath(materialPath).GetName();
+      }
       return mat;
     }
 
--- a/package/com.unity.formats.usd/Runtime/Scripts/IO/Materials/MaterialImporter.cs
+++ b/package/com.unity.formats.usd/Runtime/Scripts/IO/Materials/MaterialImporter.cs
@@ -155,6 +155,7 @@ namespace Unity.Formats.USD {
         matAdapter.ImportFromUsd();
       }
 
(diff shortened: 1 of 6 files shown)


package/com.unity.formats.usd/Runtime/Scripts/IO/Materials/MaterialImporter.cs
```diff
@@ -155,6 +155,7 @@ System.Collections.Generic.List<string> primvars
         matAdapter.ImportFromUsd();
       }
 
+      mat.name = materialPath;
```
judubu: I think I'd go with the material name instead of the full path. I'm not sure there's any value in having the full path as material name. And paths can get extremely long in the case of nested assets.

judubu (changes requested):
Looks good to me. Use the material path with the material name and it;s a go.

package/com.unity.formats.usd/Runtime/Scripts/IO/Materials/MaterialImporter.cs
vkovec: will fix. I went with the path in case there would be an issue with multiple materials of the same name being declared at different paths.

judubu: Yeah that's what I guessed. As Unity can deal with duplicated names on GO  I think we can live with that.

lucillecaillaud (approved):

judubu (approved):
👍

PR #215, opened 2020-10-27
USDU-13 add null check for scene.AccessMask

### LINKED ISSUE #174: GameObject import does not import material names (FTV-661)

vkovec:
## Purpose of this PR

**Ticket/Jira #:** USDU-13

Found while working on USDU-13. Add a null check for scene.AccessMask before calling BuildMesh. It is null when importing a usd file containing instances. 

## Testing

Manual testing:
1. Create a cube in Maya with instances
2. Export cube with instances to USD
3. In Unity, import the file with USD > Import as GameObjects
4. Before fix: NullReferenceException and instances did not import
5. After fix: no error and instances import

This PR will not affect performance.

## Overall Product Risks

**Complexity:**

Minimal

**Halo Effect:**

Minimal

## Additional information

**Note to reviewers:**

**Reminder:**


judubu (approved):
Good catch.

lucillecaillaud (approved):

PR #217, opened 2020-10-28
JIRA: USDU-57 Field-of-view animation doesn't import

clusty:
The problem was an access mask missing feature.

Upon writing the unit test I noticed that the USD recorder is not very frame accurate: the generated use files do not coincide 100% with the animation (about one frame off), thus the very large epsilons in the Assertion.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -64,3 +64,4 @@ TestProject/Logs
 TestProject/Assets/Plugins
 TestProject/Assets/Samples*
 inst-dbg/
+package/com.unity.formats.usd/TestRunnerOptions.json.meta
--- a/package/com.unity.formats.usd/Runtime/Scripts/IO/Geometry/CameraImporter.cs
+++ b/package/com.unity.formats.usd/Runtime/Scripts/IO/Geometry/CameraImporter.cs
@@ -13,6 +13,7 @@
 // limitations under the License.
 
 using UnityEngine;
+using USD.NET;
 using USD.NET.Unity;
 
 namespace Unity.Formats.USD {
@@ -25,13 +26,18 @@ namespace Unity.Formats.USD {
     /// <summary>
     /// Copy camera data from USD to Unity with the given import options.
     /// </summary>
-    public static void BuildCamera(CameraSample usdCamera,
-                                 GameObject go,
-                                 SceneImportOptions options) {
-      var cam = ImporterBase.GetOrAddComponent<Camera>(go);
-      usdCamera.CopyToCamera(cam, setTransform: false);
-      cam.nearClipPlane *= options.scale;
-      cam.farClipPlane *= options.scale;
+    public static void BuildCamera(pxr.SdfPath path,
+      CameraSample usdCamera,
+      GameObject go,
+      SceneImportOptions options, Scene scene)
+    {
+      if (scene.AccessMask == null || scene.IsPopulatingAccessMask || scene.AccessMask.Included.ContainsKey(path))
+      {
+        var cam = ImporterBase.GetOrAddComponent<Camera>(go);
+        usdCamera.CopyToCamera(cam, false);
+        cam.nearClipPlane *= options.scale;
+        cam.farClipPlane *= options.scale;
+      }
     }
   }
 }
--- a/package/com.unity.formats.usd/TestRunnerOptions.json
+++ b/package/com.unity.formats.usd/TestRunnerOptions.json
@@ -1,0 +1,3 @@
+{
+  "disableBatchMode": true
+}
--- a/package/com.unity.formats.usd/Tests/Runtime/AnimatedUSDFixture.cs
+++ b/package/com.unity.formats.usd/Tests/Runtime/AnimatedUSDFixture.cs
@@ -1,3 +1,4 @@
+#if UNITY_2019_4_OR_NEWER
 using System;
 using System.Collections;
 using System.Collections.Generic;
@@ -171,3 +172,5 @@ namespace Unity.Formats.USD.Tests
         }
     }
 }
+
+#endif
--- a/package/com.unity.formats.usd/Tests/Runtime/AnimatedUSDFixture.cs
+++ b/package/com.unity.formats.usd/Tests/Runtime/AnimatedUSDFixture.cs
@@ -44,6 +44,7 @@ namespace Unity.Formats.USD.Tests
             m_filesToDelete.Add(usdRecorderAsset.m_usdFile);
             usdRecorderAsset.m_exportRoot = new ExposedReference<GameObject> {exposedName = Guid.NewGuid().ToString()};
             director.SetReferenceValue(usdRecorderAsset.m_exportRoot.exposedName, cam);
+            Time.captureDeltaTime = 1 / timeline.editorSettings.fps;
             director.Play();
             while (director.time <= 1.1)
                 yield return null;
@@ -77,8 +78,8 @@ namespace Unity.Formats.USD.Tests
             
             director.time = 1;
             director.Evaluate();
-            Assert.That(100, Is.EqualTo(cam.transform.localPosition.x).Within(0.1));
-            Assert.That(100, Is.EqualTo(cam.fieldOfView).Within(0.1));
+            Assert.That(100, Is.EqualTo(cam.transform.localPosition.x).Within(1e-5));
+            Assert.That(100, Is.EqualTo(cam.fieldOfView).Within(1e-5));
             
         }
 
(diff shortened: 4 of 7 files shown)


package/com.unity.formats.usd/Runtime/Scripts/IO/Geometry/CameraImporter.cs
```diff
@@ -25,13 +26,18 @@ public static class CameraImporter {
...
-      cam.farClipPlane *= options.scale;
+    public static void BuildCamera(pxr.SdfPath path,
+      CameraSample usdCamera,
+      GameObject go,
+      SceneImportOptions options, Scene scene)
+    {
+      if (scene.AccessMask == null || scene.IsPopulatingAccessMask || scene.AccessMask.Included.ContainsKey(path))
+      {
+        var cam = ImporterBase.GetOrAddComponent<Camera>(go);
+        usdCamera.CopyToCamera(cam, false);
```
judubu, line 37: This will update all camera attributes whether they are animated or not. I think we should only update the attributes that are indeed dynamic (the ones in the accessMask).
What do you think?

jcowles: I think the all-or-nothing implementation is a good place to start. I would definitely profile to see the impact of making it more granular -- I'm not opposed to going more granular, but it's good practice to profile any time you make a complicating change for performance.

.gitignore
```diff
@@ -64,3 +64,4 @@ TestProject/Logs
 TestProject/Assets/Plugins
 TestProject/Assets/Samples*
 inst-dbg/
+package/com.unity.formats.usd/TestRunnerOptions.json.meta
```
jcowles, line 67: Why is this needed?
(any time .meta is git ignored, my spidy sense starts to tingle)

package/com.unity.formats.usd/TestRunnerOptions.json
```diff
@@ -0,0 +1,3 @@
+{
+  "disableBatchMode": true
```
jcowles, line 2: Just for my own understanding, why was this needed?
It sucks that we have no way to document this in the json :/

package/com.unity.formats.usd/Tests/Runtime/AnimatedUSDFixture.cs
```diff
@@ -0,0 +1,173 @@
...
+using NUnit.Framework;
+using UnityEditor;
+using UnityEngine;
+using UnityEngine.Playables;
+using UnityEngine.TestTools;
+using UnityEngine.Timeline;
+using USD.NET;
+using Object = UnityEngine.Object;
+
+namespace Unity.Formats.USD.Tests
```
jcowles, line 15: YAY TESTS!

package/com.unity.formats.usd/Tests/Runtime/AnimatedUSDFixture.cs
```diff
@@ -0,0 +1,173 @@
...
+            director.Evaluate();
+
+            Assert.That(50, Is.EqualTo(cam.transform.localPosition.x).Within(1e-5));
+            Assert.That(50, Is.EqualTo(cam.fieldOfView).Within(1e-5));
+            
+            director.time = 1;
+            director.Evaluate();
+            Assert.That(100, Is.EqualTo(cam.transform.localPosition.x).Within(1e-5));
+            Assert.That(100, Is.EqualTo(cam.fieldOfView).Within(1e-5));
+
```
jcowles, line 83: Extra blank space

package/com.unity.formats.usd/Tests/Runtime/AnimatedUSDFixture.cs
```diff
@@ -0,0 +1,173 @@
...
+        {
+            InitUsd.Initialize();
+            
+            var curve = AnimationCurve.Linear(0, 50, 1, 100);
+            var clip = new AnimationClip();
+            clip.SetCurve("", typeof(Transform), "localPosition.x", curve);
+            clip.SetCurve("", typeof(Camera), "field of view", curve);
+
+            CreateTimeline(out var director, out var timeline);
+
```
jcowles, line 33: I think a comment of what the test setup is actually doing would be helpful

package/com.unity.formats.usd/Tests/Runtime/AnimatedUSDFixture.cs
```diff
@@ -0,0 +1,173 @@
...
+            yield return null;
+            var stage = pxr.UsdStage.Open(usdRecorderAsset.m_usdFile, pxr.UsdStage.InitialLoadSet.LoadAll);
+            var scene =  Scene.Open(stage);
+            m_usdRoot = ImportSceneAsGameObject(scene).GetComponent<UsdAsset>();
+            scene.Close();
+            m_usdRoot.name = "Potato";
+        }
+
+        [Test]
+        public void TestCamAnimation()
```
jcowles, line 61: Similarly, I think a short comment of what this is testing and how it's testing it would be helpful here too

package/com.unity.formats.usd/Tests/Runtime/AnimatedUSDFixture.cs
```diff
@@ -0,0 +1,173 @@
...
+            var children = scene.Stage.GetPseudoRoot().GetChildren().ToList();
+            if (children.Count == 1) {
+                return children[0].GetPath();
+            }
+
+            // Otherwise there are 0 or many root prims, in this case the best option is to reference
+            // them all, to avoid confusion.
+            return pxr.SdfPath.AbsoluteRootPath();
+        }
+        static GameObject UsdToGameObject(GameObject parent,
```
jcowles, line 153: Missing new/blank lines between these last few functions

package/com.unity.formats.usd/Tests/Runtime/AnimatedUSDFixture.cs
```diff
@@ -0,0 +1,173 @@
+using System;
```
jcowles, line 1: This file is missing the copyright notice

jcowles (commented):
This is awesome, thanks Vlad!

jcowles:
Might also want to add a note to changelog.md

.gitignore
clusty: This is a strange one: this file is a part of the package, but ignored by upm-ci pack. If I were to add a meta file, you'd fail validation because you'd have a meta without the corresponding file.

package/com.unity.formats.usd/Runtime/Scripts/IO/Geometry/CameraImporter.cs
judubu: After discussing with Vlad, given we are creating a GfCamera from all cam attributes anyway, there's no direct way to go component by component beside duplicating the GfCamera code. So let's stick to the all-or-nothing approach

package/com.unity.formats.usd/TestRunnerOptions.json
clusty: Tests that rely on the rendering need a valid gameView open (to generate renders).
By default upm-ci runs runtime tests in batch mode (no rendering) causing tests to hang or straight up not render. This runs unity with a HEAD :) (probably a tad slower than batch mode).

cguer (changes requested):
I'm getting a weird result when playing the animated FOV in Unity. The FOV is properly set when first importing the camera but the values go way off when I play the animation.

This is what I have in my USD file (in the attached [camera_animatedFOV.zip](https://github.com/Unity-Technologies/usd-unity-sdk/files/5466024/camera_animatedFOV.zip))
```
float focalLength.timeSamples = {
    1: 35,
    10: 10,
}
float horizontalAperture.timeSamples = {
    1: 36,
}
float verticalAperture.timeSamples = {
    1: 24,
}
```

It should slowly "zoom out" but the fov becomes `1e-05` after two frames and the clipping planes are reset to zero.
usd_animatedfov

I get the correct result when importing the same file in USDview, Maya, and Houdini

package/com.unity.formats.usd/Tests/Runtime/AnimatedUSDFixture.cs
judubu: What's Unity policy regarding copyrights? Is there anything specific about this repo because of the Apache license?

clusty:
Closing. The issue is deeper inside lib-use-net (properties with only one sample are not deserialized correctly)

PR #219, opened 2020-10-28
Fix USD refresh not always working in Prefab stage (USDU-128)

thomas-tu:
## Purpose of this PR

**Ticket/Jira #:** USDU-128

After a USD has been loaded into a Prefab, going in Prefab Stage and refreshing the UsdAsset with Payload Policy set to Load All doesn't work.

Turns out, the code path that is being used for prefab doesn't unreference the scene before updating the data. It makes the logic re-use the cache data instead of re-opening the scene with a LoadAll policy.

## Testing

**Functional Testing status:** Manuals tests identical as the reproduction steps of the issue.

## Overall Product Risks

**Complexity:** Low

**Halo Effect:** Low

## Additional information

**Note to reviewers:**

**Reminder:**

### DIFF
--- a/package/com.unity.formats.usd/Runtime/Scripts/Behaviors/UsdAsset.cs
+++ b/package/com.unity.formats.usd/Runtime/Scripts/Behaviors/UsdAsset.cs
@@ -501,6 +501,11 @@ namespace Unity.Formats.USD {
         if (options.forceRebuild) {
           DestroyAllImportedObjects();
         }
+
+        m_lastScene = null;
+        m_lastPrimMap = null;
+        m_lastAccessMask = null;
+
         SceneImporter.ImportUsd(root, GetScene(), new PrimMap(), options);
 
 #if UNITY_EDITOR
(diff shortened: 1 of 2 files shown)


judubu (approved):
👍

vkovec (approved):

package/com.unity.formats.usd/Runtime/Scripts/Behaviors/UsdAsset.cs
```diff
@@ -443,6 +442,16 @@ public string usdFullPath
       if (!comp) { return; }
       Component.DestroyImmediate(comp);
     }
+    
+    /// <summary>
+    /// Clear internal data.
+    /// Call to <see cref="GetScene">GetScene()</see> to update them with the latest USD data.
+    /// </summary>
+    private void ClearLastData() {
+      m_lastScene = null;
```
jcowles, line 452: You and @lucillecaillaud should talk, I think her change to just call stage.Reload() may be a better option, though I'm not sure now if her change also requires clearing the prim map and access mask.

package/com.unity.formats.usd/Runtime/Scripts/Behaviors/UsdAsset.cs
```diff
@@ -501,6 +511,8 @@ public string usdFullPath
         if (options.forceRebuild) {
           DestroyAllImportedObjects();
         }
+
+        ClearLastData();
```
thomas-tu, line 515: Protect this, it should happen only when there is a change of load policy.

thomas-tu:
Closing it as some improvements can be done.

PR #220, opened 2020-10-29
Fix corrupted scene primMap with invalid prim (USDU-60, #158)

### LINKED ISSUE #23: USD.NET.Scene throws exceptions for non-exiting prims (FTV-210)
When a prim doesn't exist, USD.NET.Scene attempts to cache a null UsdPrim, which will cause it to be cached multiple times, which ultimately throws Dictionary exceptions that the entry already exists.

While this is a user/coding error, it is also a confusing error message.

### LINKED ISSUE #158: Calling Scene.Read on a non-existent prim causes a crash on trying to subsequently write to that prim (FTV-659)
Hi

If you call Scene.Read on a non-existent prim, the Scene.GetUsdPrim method adds a new invalid prim to the prim cache. If you subsequently try to write that prim, the prim is not created because the invalid one is already there, and the serialization code will crash with a USD FATAL ERROR null prim message.

lucillecaillaud:
## Purpose of this PR

**Ticket/Jira #:** [USDU-60](https://jira.unity3d.com/browse/USDU-60)

Getting a prim at a path that doesn't exists returns and invalid prim.
In the Scene.GetUsdPrim function (called by Scene.Read), this invalid prim was added to the scene primMap.
When trying to write (scene.Write) the prim at this path it was causing error because the invalid prim was already written.

Now, prim are added to the primMap only if they exist and are valid.

## Testing

Added the SceneTests.cas test class with two tests:
* One that test to read then write a valid prim
* One that read an invalid prim then write it (this test should make Unity crash prior to this changes)

N/A

## Overall Product Risks

**Complexity:** Minimal

**Halo Effect:** Minimal

## Additional information

**Note to reviewers:**

**Reminder:**


jcowles:
I believe this also fixes Issue #23

src/USD.NET/serialization/Scene.cs
```diff
@@ -745,10 +745,12 @@ public enum UpAxes {
...
       lock(m_stageLock) {
-      if (!m_primMap.TryGetValue(path, out prim) || !prim.IsValid()) {
-        prim = Stage.GetPrimAtPath(path);
-        m_primMap[path] = prim;
-      }
+        if (!m_primMap.TryGetValue(path, out prim) || !prim.IsValid()) {
+          prim = Stage.GetPrimAtPath(path);
+          if (prim.IsValid()) {
+            m_primMap[path] = prim;
+          }
```
jcowles, line 752: If the prim isn't valid (after calling GetPrimAtPath) and the path **does** exist in the map, should the path be cleared here to keep the primMap accurate?

jcowles (commented):
Awesome :)

vkovec (approved):
I think you accidentally commited the dll and pdb, otherwise looks good to me

lucillecaillaud:
> I think you accidentally commited the dll and pdb, otherwise looks good to me

It's not an accident. My change is in USD.NET, so I rebuilt this dll, that's why I'm pushing the updated version to the package Plugins.

src/USD.NET/serialization/Scene.cs
lucillecaillaud: That's a good question... I don't know ^^. What would be the case where it happens?... If a prim is removed on disk maybe?? Then, when refreshing... the primMap wouldn't be accurate?

jcowles: Yeah, that's what I was thinking -- like a file is regenerated and some prims are removed, or some flavor of that.

I don't think this is a /bug/ but I think it would be a good axiom to enforce that the prim map is always either pointing to a valid prim or has no entry.

jcowles: (It should be super easy to test this with an ascii file, just delete the prim and see what happens on reload at that point, but as I read the code, it looks like an entry will be left in the prim map)

judubu: @jcowles What is the benefit of the primMap here? Stage.GetPrimAtPath is a sdf operation and will be cached anyway so it should be fast. 
If we keep a local cache in Scene I can't see a reliable way to ensure the cache stays in sync while we expose the stage as public. We could provide our own Scene.RemovePrim and Scene.Reload but nothing prevents the users to call scene.Stage.RemovePrim or scene.Stage.Reload ...

package/com.unity.formats.usd/Tests/Editor/SceneTests.cs
```diff
@@ -0,0 +1,106 @@
+﻿using NUnit.Framework;
```
judubu: Thanks for adding the test @lucillecaillaud That's what I had in mind.
But thinking about it, this test should really go with the other USD.NET tests. I'll move it there.

judubu (changes requested):
Good job @lucillecaillaud 
I'll move the test with the other USD.NET test though.
Also wondering about if the primMap is a necessity. Waiting for @jcowles to comment.

src/USD.NET/serialization/Scene.cs
jcowles: I believe the original motivation was to avoid the C++/C# roundtrip just to get a prim, though it seems silly now.

I wonder if we can just kill it entirely -- given that this requires grabbing a lock, seems like that would be nice.

judubu (approved):

src/USD.NET/serialization/Scene.cs
judubu: Created a ticket for this one: https://jira.unity3d.com/browse/USDU-135

PR #221, opened 2020-10-29
USDU-46 create sparse timesamples

vkovec:
## Purpose of this PR

**Ticket/Jira #:** USDU-46

Fix redundant timesamples created every frame for objects that are not animated, or whose animation is not changing.

## Testing

Added a unit test for recording a simple hierarchy to USD from the timeline and verifying that only the animated objects contain timesamples in the USD file.

Manual Test 1:
1. Create a cube and a sphere parented under it (or test with a different file that has at least one unanimated object)
2. Create a timeline and add the sphere to it
3. Animate the sphere but not the cube
4. Add a USD recorder to the timeline
5. Set the USD recorder export root to be the cube
6. Click play to record
7. Stop playing
8. Inspect recorded file, check that there are no timesamples for the cube

Manual Test 2:
1. Run Manual Test 1 with the HumanFemale walking asset
2. Import into both Unity and Maya -> check that it imports upright
Note: there are some issues with the animation, though I think they were present before the fix as well.

[recording_person_before_and_after_fix.zip](https://github.com/Unity-Technologies/usd-unity-sdk/files/5461249/recording_person_before_and_after_fix.zip)

Manual Test 3:
1. Run the Export Mesh sample
2. Check that it gives the correct result

Profiled export of HumanFemale walking asset with and without change. Gave similar results in profiler.

## Overall Product Risks

**Complexity:**

Low

**Halo Effect:**

Medium - changes the way attribute values are written (through the UsdUtilsSparseAttrValueWriter as opposed to written directly with attr.SetValue()).

## Additional information

**Note to reviewers:**

**Reminder:**


package/com.unity.formats.usd/Runtime/Scripts/IO/Skel/SkeletonExporter.cs
```diff
@@ -101,6 +101,18 @@ public static class SkeletonExporter {
         sample.extent.center = UnityTypeConverter.ChangeBasis(sample.extent.center);
       }
 
+      // Convert the transform
```
vkovec, line 104: Transform timesamples were being exported zeroed out for the Skeleton Root.

package/com.unity.formats.usd/Tests/Runtime/AnimationRecordingTests.cs
```diff
@@ -0,0 +1,109 @@
+﻿using System;
```
vkovec, line 1: Test based off https://github.com/Unity-Technologies/usd-unity-sdk/pull/217/files

src/USD.NET/serialization/UsdIo.cs
```diff
@@ -476,7 +478,7 @@ public class UsdIo {
         } else if (Reflect.IsFusedDisplayColor(memberInfo)) {
           pxr.UsdCs.SetFusedDisplayColor(prim, vtValue, time);
```
judubu, line 483: I suppose we should also do it for the FusedDisplayColor if we don't want time samples on those. It's directly in the swig bindings though so I'll create a Jira issue for this one and do it with a batch of other bindings update.

package/com.unity.formats.usd/Tests/Runtime/AnimationRecordingTests.cs
judubu: Thanks for adding a test. I realize now that the fix is in USD.NET so the test should actually go in the USD.NET tests. I'll move the test there myself.

package/com.unity.formats.usd/Runtime/Scripts/IO/Skel/SkeletonExporter.cs
judubu: :+1:

judubu (approved):
Looks good.

cguer (changes requested):
If I understood this change correctly, every contiguous `timeSamples` with identical values should get "compressed". During testing, I noticed that only `timeSamples` with values matching the attribute's default value actually get "compressed". For example:

Here is the animation within Unity. The object's position is `(333, 333, 333)` by default, then goes down to `(0,0,0)` and back up to `(333, 333, 333)`.
Annotation 2020-11-18 104631

The animation above produces the following USD file using the old method `attr.Set` (Note: Z position is now negative due to handedness conversion)
<pre>
matrix4d xformOp:transform = ( (1, 0, 0, 0), (0, 1, 0, 0), (0, 0, 1, 0), (333, 333, -333, 1) )
matrix4d xformOp:transform.timeSamples = {
    0: ( (1, 0, 0, 0), (0, 1, 0, 0), (0, 0, 1, 0), (333, 333, -333, 1) ),
    1.0000000521540642: ( (1, 0, 0, 0), (0, 1, 0, 0), (0, 0, 1, 0), (333, 333, -333, 1) ),
    2.0000001043081284: ( (1, 0, 0, 0), (0, 1, 0, 0), (0, 0, 1, 0), (333, 333, -333, 1) ),
    3.0000001564621925: ( (1, 0, 0, 0), (0, 1, 0, 0), (0, 0, 1, 0), (333, 333, -333, 1) ),
    4.000000208616257: ( (1, 0, 0, 0), (0, 1, 0, 0), (0, 0, 1, 0), (333, 333, -333, 1) ),
    <b>5.000000260770321: ( (1, 0, 0, 0), (0, 1, 0, 0), (0, 0, 1, 0), (0, 0, 0, 1) ),
    6.000000312924385: ( (1, 0, 0, 0), (0, 1, 0, 0), (0, 0, 1, 0), (0, 0, 0, 1) ),
    7.000000365078449: ( (1, 0, 0, 0), (0, 1, 0, 0), (0, 0, 1, 0), (0, 0, 0, 1) ),
    8.000000417232513: ( (1, 0, 0, 0), (0, 1, 0, 0), (0, 0, 1, 0), (0, 0, 0, 1) ),
    9.000000469386578: ( (1, 0, 0, 0), (0, 1, 0, 0), (0, 0, 1, 0), (0, 0, 0, 1) ),</b>
    10.000000521540642: ( (1, 0, 0, 0), (0, 1, 0, 0), (0, 0, 1, 0), (333, 333, -333, 1) ),
    11.000000573694706: ( (1, 0, 0, 0), (0, 1, 0, 0), (0, 0, 1, 0), (333, 333, -333, 1) ),
    12.00000062584877: ( (1, 0, 0, 0), (0, 1, 0, 0), (0, 0, 1, 0), (333, 333, -333, 1) ),
    13.000000678002834: ( (1, 0, 0, 0), (0, 1, 0, 0), (0, 0, 1, 0), (333, 333, -333, 1) ),
    14.000000730156898: ( (1, 0, 0, 0), (0, 1, 0, 0), (0, 0, 1, 0), (333, 333, -333, 1) ),
}
</pre>

And here is the result from the current PR using `pxr.UsdUtilsSparseAttrValueWriter()`. Notice that only the timeSample that have exactly the same position as the default value get "compressed". All other values, even if they have contiguous identical samples, are untouched:
```
matrix4d xformOp:transform = ( (1, 0, 0, 0), (0, 1, 0, 0), (0, 0, 1, 0), (333, 333, -333, 1) )
matrix4d xformOp:transform.timeSamples = {
    5.000000260770321: ( (1, 0, 0, 0), (0, 1, 0, 0), (0, 0, 1, 0), (0, 0, 0, 1) ),
    6.000000312924385: ( (1, 0, 0, 0), (0, 1, 0, 0), (0, 0, 1, 0), (0, 0, 0, 1) ),
    7.000000365078449: ( (1, 0, 0, 0), (0, 1, 0, 0), (0, 0, 1, 0), (0, 0, 0, 1) ),
    8.000000417232513: ( (1, 0, 0, 0), (0, 1, 0, 0), (0, 0, 1, 0), (0, 0, 0, 1) ),
    9.000000469386578: ( (1, 0, 0, 0), (0, 1, 0, 0), (0, 0, 1, 0), (0, 0, 0, 1) ),
}
```
----
⚠️ Warning ⚠️
There is currently a bug in the USD Recorder that discards/ignores the first 1-2 seconds of samples. In the meantime, I would suggest doing one of the following to workaround the issue during testing:

- Disable `Play on Awake` for the Timeline's Playable Directory. Enter Play mode, _then_ manually play the Timeline
- Move the start of the USDrecorder clip so that it starts 1-2 seconds after the beginning of Timeline playback
- Add a (regular) Recorder Track and clip. Having a Recorder track clip running at the same time better controls the GameTime (no missing frame at start and increased frame accuracy)

judubu (approved):
Thanks for the fix. I was not sure how to use the sparseWriter but it make sense now. :+1:
Note that I started to refactor the USD.NET tests in a separate task and I will update the attribute writing tests to test for sparse writing and remove the tests from this PR at that moment.

cguer (approved):
QA approved 👍!

Tested on:
- 2019.4.13f1 (Windows only)

PR #222, opened 2020-10-29
Exporting transform overrides should export usda files (USDU-124)

mfe:
## Purpose of this PR

**USDU-124**

Right now, the Transform Export Override is exporting a .usda.usd file that actually is an usdc.

## Testing

Repro:
* Import a usd file as GO
* Then USD > Export Transform Overrides

Before => in the saving window, file name extension is .usda.usd. If you keep it like that, the save USD file is an USDc
Now => file name extension and content is usda

**Complexity:**
Minimal 

**Halo Effect:**
Minimal 

## Additional information

**Note to reviewers:**
It's a one-liner :D

**Reminder:**
- [ X ] Add entry in CHANGELOG.md _(if applicable)_


judubu (approved):
👍

package/com.unity.formats.usd/Editor/Scripts/Behaviors/UsdMenu.cs
```diff
@@ -74,7 +74,7 @@ public class UsdMenu : MonoBehaviour {
     [MenuItem("USD/Export Transform Overrides", priority = 50)]
     static public void MenuExportTransforms() {
       var root = Selection.activeGameObject.GetComponentInParent<UsdAsset>();
-      var overs = InitForSave(Path.GetFileNameWithoutExtension(root.usdFullPath) + "_overs.usda");
+      var overs = InitForSave(Path.GetFileNameWithoutExtension(root.usdFullPath) + "_overs", fileExtension:"usda");
```
jcowles: doh, thanks for the fix!

jcowles (approved):

package/com.unity.formats.usd/Editor/Scripts/Behaviors/UsdMenu.cs
thomas-tu: Does it have to be in `usda`? We can maybe suggest all compatible formats if you give: `"usd,usdc,usda"`. The save panel will then show as able to handle these formats. It would also make things easier if someone wants to overwrite an existing file.

Going further... I just saw what could be inconsistent:
`string path = EditorUtility.OpenFilePanel("Import USD File", "", "usd,usda,usdc,abc");`

`var overs = InitForSave(Path.GetFileNameWithoutExtension(root.usdFullPath) + "_overs", fileExtension:"usda");`

`public static Scene InitForSave(string defaultName, string fileExtension = "usd")`

` static void ExportSelected(BasisTransformation basisTransform,
                               string fileExtension = "usd",
                               bool exportMonoBehaviours = false)`

If I may, I would suggest to offer `usd,usda,usdc` by default for `InitForSave` and `ExportSelected()` and override it if there's a strong need of something different.

cguer: I agree with Thomas, is there a particular reason we are limiting the export format to `usda`?

cguer (commented):
I'm somehow unable to reproduce the original issue (double-extension in file name) using USD 1.0.3.

These are the default settings I get in the export window on Windows. The generated file on disk is `moving_cube_overs.usda` and the content is ASCII.
Annotation 2020-11-11 102639


On OSX, the file exports as `.usd`, the content is binary, and I get a warning if I try to manually change the extension in the export window.

Do I need a specific version of the Editor to repro?

package/com.unity.formats.usd/Editor/Scripts/Behaviors/UsdMenu.cs
judubu: You're right, there's no reason to force ascii over binary. Let's do as Thomas proposed. 
Ideally we should offer an option to select if the usd extension should be ascii or binary as both are valid. But let's keep that for when we refactor the exporter to expose export options.

judubu: Just did it in 6f9f91f

thomas-tu (approved):

cguer (approved):
QA approved 👍!

I'm still somehow unable to reproduce the original issue (I have video proof) but the changes from this PR look good!

Tested on:
- 2019.4.13f (Windows only)

PR #223, opened 2020-10-30
(USDU-9) Left handed orientation not handled correctly

### LINKED ISSUE #76: MeshImport: orientation="leftHanded" not handled correctly (FTV-204)
For example, HullBrush exports from Tilt Brush are not imported correctly, resulting in a mesh which is inside out.

image

[HullBrush.zip](https://github.com/Unity-Technologies/usd-unity-sdk/files/3050631/HullBrush.zip)

vkovec:
## Purpose of this PR

**Ticket/Jira #:** USDU-9

**Edit: After investigation this is an error with this specific USD file and not the import code. Changes reverted and added a unit test**

If a mesh is lefthanded orientation in the USD file, then it will be partially converted despite being lefthanded already. Fixed to check if mesh is lefthanded earlier and convert accordingly.

Before fix (compared to FBX version exported from Maya, on right):
image

After fix (compared to FBX version exported from Maya, on right):
image

## Testing

Added an automated test to compare left vs right handed import.

Manual Test:
1. Create a left handed mesh in USD (or use the one linked to the issue)
2. In Unity, run USD > Import as GameObjects
3. In the Unity Asset, under "Advanced", change the "Change Handedness" settings, refresh and check that result is as expected.

Should not affect performance.

## Overall Product Risks

**Complexity:**

Low

**Halo Effect:**

Low

## Additional information

**Note to reviewers:**

**Reminder:**


judubu (approved):
Damn, I looked at this code so many times before, and still I missed it!
Good job.

package/com.unity.formats.usd/Tests/Editor/EditorTests.cs
```diff
@@ -67,6 +75,29 @@ public void LoadAsFbxCompareMeshTransforms()
             Assert.AreApproximatelyEqual(usdMeshTr.localRotation.y,fbxMeshTr.localRotation.y);
             Assert.AreApproximatelyEqual(usdMeshTr.localRotation.z,fbxMeshTr.localRotation.z);
         }
+
+        [Test]
+        public void TestLeftHandedUsdImport()
```
vkovec: there is probably a better way to test this, adding this for now to check that the same cube is handled differently whether it is right or left handed.

jcowles:
This should also fix issue #76 which has a repro.

At first glance what you have seems correct, but I want to read over it more carefully later today

This is awesome!

package/com.unity.formats.usd/Tests/Editor/EditorTests.cs
```diff
@@ -4,34 +4,42 @@
...
 using Assert = UnityEngine.Assertions.Assert;
+using System.Collections;
 
 namespace Unity.Formats.USD.Tests
 {
     public class FBXHandednessModeTests
     {
         const string fbxGUID = "86a597c63449d2541b7587ff90e75d91"; // GUID of withCamera.fbx
         const string usdGUID = "f377c4260fb216d4dbe2f6e4d67091b5"; // GUID of withCamera.usd
+        const string leftHandedUsdGUID = "25388fbb44b88c448a40ef8eeb0c3ca2"; // GUID of withCamera_leftHanded.usd
```
jcowles: If this class were a ScriptableObject, you could actually just bind these assets, rather than hard-coding the GUID, I think (I have not tried this, it was just a passing thought)

package/com.unity.formats.usd/Runtime/Scripts/IO/Geometry/MeshImporter.cs
```diff
@@ -284,13 +284,16 @@ public class GeometrySubsets {
       Material mat = renderer.sharedMaterial;
       bool changeHandedness = options.changeHandedness == BasisTransformation.SlowAndSafe ||
                               options.changeHandedness == BasisTransformation.SlowAndSafeAsFBX;
+      // check if the handedness needs to change
+      bool isLeftHanded = usdMesh.orientation == Orientation.LeftHanded;
+      var handednessChanged = changeHandedness && !isLeftHanded || !changeHandedness && isLeftHanded;
```
jcowles: For readability, rather than making a new variable, could we just reuse `changeHandedness`?
I would also even kill `isLeftHanded` and just accumulate that into `changeHandedness` as well.

Otherwise when reading this code, you will load three variables into your brain, only to realize that two of them are unused after this line.

jcowles (commented):
This is awesome, thanks for the fix!

package/com.unity.formats.usd/Tests/Editor/EditorTests.cs
```diff
@@ -67,6 +75,29 @@ public void LoadAsFbxCompareMeshTransforms()
...
+            Assert.IsNotNull(leftHandedUsdCube);
+
+            var cubeMesh = usdCube.GetComponent<MeshFilter>().sharedMesh;
+            var leftHandedCubeMesh = leftHandedUsdCube.GetComponent<MeshFilter>().sharedMesh;
+            
+            // Since the two files are different handedness, the vertices, triangles, and normals
+            // will be different.
+            NUnit.Framework.Assert.That(cubeMesh.vertices, Is.Not.EqualTo(leftHandedCubeMesh.vertices));
+            NUnit.Framework.Assert.That(cubeMesh.triangles, Is.Not.EqualTo(leftHandedCubeMesh.triangles));
+            NUnit.Framework.Assert.That(cubeMesh.normals, Is.Not.EqualTo(leftHandedCubeMesh.normals));
```
thomas-tu: Is it enough to compare the data of two import results? Shouldn't we also test the generated mesh against a baked version to ensure the result we get is correct?

package/com.unity.formats.usd/Tests/Editor/EditorTests.cs
vkovec: cool it worked, thanks!

package/com.unity.formats.usd/Tests/Editor/EditorTests.cs
jcowles: This is actually a really good point -- especially with a symmetrical mesh like a Cube, you might even be getting identical results, but the floating point error alone might make them look different.

Does `Is.Not.EqualTo` use a fuzzy/epsilon comparison? If so, is the epsilon small enough to work here?

jcowles: I do like NOT using a baked/golden result though, because it makes the test more robust to other changes and avoids needing to ratchet a baseline.

package/com.unity.formats.usd/Runtime/Scripts/IO/Geometry/MeshImporter.cs
```diff
@@ -284,6 +284,9 @@ public class GeometrySubsets {
       Material mat = renderer.sharedMaterial;
       bool changeHandedness = options.changeHandedness == BasisTransformation.SlowAndSafe ||
                               options.changeHandedness == BasisTransformation.SlowAndSafeAsFBX;
+      // Check if the handedness needs to change.
+      changeHandedness = changeHandedness && (usdMesh.orientation != Orientation.LeftHanded) ||
+                !changeHandedness && (usdMesh.orientation == Orientation.LeftHanded);
```
jcowles: When mixing && and ||, you should add parentheses to be sure the logic works the way you expect and to make it clear to future readers

jcowles (approved):
Nice, LGTM!

jcowles:
I approved this, leaving it up to you how to handle the conversation about testing

package/com.unity.formats.usd/Tests/Editor/FBXHandednessModeTests.cs
```diff
@@ -0,0 +1,112 @@
+﻿using NUnit.Framework;
```
jcowles: I just noticed, no copyright disclaimer was added to this file

package/com.unity.formats.usd/Tests/Editor/EditorTests.cs
vkovec: It should call Equals(), however, looking at the [Vector3 documentation](https://docs.unity3d.com/ScriptReference/Vector3.Equals.html) it looks like == is safer than Equals for approximate comparison, so I'll change this to be more explicit.

jcowles:
Thanks for incorporating all the feedback on this one!
And nice work :)

cguer (commented):
I'm always a bit confused about what the "truth" is when reviewing handedness changes but I noticed one possibly weird behavior testing the assets from the case.

Note: The images below are using the FBX importer in 2019.4, NOT with the new `Bake Axis Conversion` in 2020.1+. (should I?)

**USD version 1.0.3 (last release)**
When the `Change Handedness` parameter is set to `Slow And Safe As FBX`, the USD geometry matches exactly the FBX geometry
usd_vs_fbx_packman

**USD version 1.0.4 (current PR)**
There appears to be no difference between `Slow And Safe As FBX` and `Slow And Safe`.

Is this the expected behavior?
usd_vs_fbx_pr

vkovec:
Did some testing with the following software:
- usdview
- Maya
- Cinema4D (same axis system as Unity)

Cinema4D and Maya both seem to export right handed USD files.

I had a hard time finding software to generate left handed USD files, so instead tried making a right handed version of the original hull brush file.
To generate the right handed version, I simply edited all the orientation values in the file to "rightHanded".
Then I did some tests to see how it would be imported compared to the leftHanded version in different software.

All three software listed above imported both the left handed and right handed version in the same way.

In the Maya USD importer, it appears that the orientation is only considered if an object is not doubleSided.
Therefore, I set doubleSided = 0 in both right and left handed files. This time they imported slightly different in usdview and Maya.
(right handed file on left, left handed file on right)
image

In Unity, the right handed version matches the FBX version. See the following two images:

In the below images, the following 4 files were imported into a scene:
1. an FBX version of the model created by importing the USD into Maya, then exporting as FBX, imported with bake axis conversion disabled.
2. the original left handed USD. Imported with "Slow and Safe" conversion.
3. the original USD, edited to change all instances of `uniform token orientation = "leftHanded"` to `uniform token orientation = "rightHanded"`. Imported with "Slow and Safe" conversion.
4. Same as 3) but imported with "Slow and Safe as FBX" conversion.

With USD package using fix from this PR:
image

with USD package version 1.0.3 from package manager:
image

* Note: the FBX imported with bake axis conversion enabled matches 3) in the images.

Based on this information, I may have misunderstood left vs right handed orientation in USD, and the issue seems to be with the file rather than the importer. The file appears to actually be a right handed file.

jcowles:
Perhaps this is more helpful, here is a simple USD file with a right and left handed quad:

```
#usda 1.0

def Xform "World" {

    def Camera "MainCam" {
        double3 xformOp:translate = (0, 0, 10)
        uniform token[] xformOpOrder = [ "xformOp:translate" ] 
        float2 clippingRange = (1, 20)
    }

    def Mesh "RightHanded" {
        token orientation = "rightHanded"
        int[] faceVertexCounts = [4]
        int[] faceVertexIndices = [0, 1, 2, 3]
        point3f[] points = [(0,0,0),
                            (1,0,0),
                            (1,1,0),
                            (0,1,0)]
    }

    def Mesh "LeftHanded" {
        token orientation = "leftHanded"
        int[] faceVertexCounts = [4]
        int[] faceVertexIndices = [0, 3, 2, 1 ]
        point3f[] points = [(-2,0,0),
                            (-1,0,0),
                            (-1,1,0),
                            (-2,1,0)]
    }
}
```

And this is what it looks like in usdview:
image

And through the camera:
image

jcowles:
In the current version of the USD Unity SDK, I get:

image

Which looks correct to me.

I also noticed that the latest revision of this change is effectively a no-op now, right? There is no functional change from `stable`.

I think Houdini will export left-handed USD files, IIRC.

jcowles:
So maybe there was never a bug? *facepalm*

vkovec:
I agree it looks like this was never actually a bug. Here is another example with a left handed example file from Houdini:

In Houdini:
image

In usdview:
image

In Maya (not sure why it imports black, otherwise looks correct):
image

In Unity with USD 1.0.3 (package manager):
image

In Unity with this branch:
image

In the last one the normals look off in some areas, in addition to facing the wrong way.

judubu:
So it seems the conclusion is there never was a bug, it's the tilt brush asset that was wrong. Sorry @vkovec for the time wasted, and thanks to everybody that spent time looking into the issue. 
If nobody comment I'll close the ticket.

jcowles:
It might be nice to keep the Unit test from this PR, but revert the functional changes.

vkovec:
@judubu @jcowles sorry for the confusion, I have reverted the changes and fixed up the unit tests to match the correct functionality.

judubu:
Awesome. Thanks @vkovec  Ill merge as soon as Yamato turns green.

PR #228, opened 2020-11-26
Add EditorPrefs for payload policy and material import mode

judubu:
## Purpose of this PR

**Ticket/Jira #:** https://jira.unity3d.com/browse/USDU-141

Default values for the UsdAsset were set to not load the payloads and not load the materials. It gives the feeling the importer is broken and it's not clear to a new user which option needs to be changed to load the full asset.

This PR changes the default values  to load the payloads and the materials, and adds those settings as EditorPrefs. 
The settings are accessible in the USD/Preferences sub menu

usd_prefs_menu

Note that the default values of the UsdAsset component were not changed but the user have full control over them.

## Testing

Loading the usd kitchen shows that the payload policy is respected
Loading the usd female skeleton show that the material import mode is respected
Restarting Unity after changing the prefs shows that the prefs are registered
I also checked that the prefs were properly set in the windows registry

## Overall Product Risks

**Complexity:**
Minimal

**Halo Effect:**
Minimal

## Additional information

**Note to reviewers:**

I also reformated the code to respect Unity code style. I'll do that progressively on all files I modify.
The functional changes are in the first commit.

**Reminder:**

### DIFF
--- a/package/com.unity.formats.usd/Editor/Scripts/Behaviors/UsdMenu.cs
+++ b/package/com.unity.formats.usd/Editor/Scripts/Behaviors/UsdMenu.cs
@@ -20,8 +20,14 @@ using USD.NET;
 using USD.NET.Unity;
 
 namespace Unity.Formats.USD {
- 
-  public class UsdMenu : MonoBehaviour {
+
+  public class UsdMenu : MonoBehaviour
+  {
+      const string PayloadPolicyPrefName = "USD.LoadAllPayloads";
+      const string MaterialPolicyPrefName = "USD.LoadPreviewSurface";
+
+      const string UsdPrefPayloadMenuName = "USD/Preferences/Load all payloads";
+      const string UsdPrefMaterialMenuName = "USD/Preferences/Load preview surface";
 
     public static Scene InitForSave(string defaultName, string fileExtension = "usd,usda,usdc") {
       var filePath = EditorUtility.SaveFilePanel("Export USD File", "", defaultName, fileExtension);
@@ -54,7 +60,10 @@ namespace Unity.Formats.USD {
         return null;
       }
       InitUsd.Initialize();
-      var stage = pxr.UsdStage.Open(path, pxr.UsdStage.InitialLoadSet.LoadNone);
+      var loadSet = EditorPrefs.GetBool(PayloadPolicyPrefName, true)
+          ? pxr.UsdStage.InitialLoadSet.LoadAll
+          : pxr.UsdStage.InitialLoadSet.LoadNone;
+      var stage = pxr.UsdStage.Open(path, loadSet);
       return Scene.Open(stage);
     }
 
@@ -192,7 +201,38 @@ namespace Unity.Formats.USD {
       }
     }
 
-    [MenuItem("USD/Import as GameObjects", priority = 0)]
+
+    [MenuItem(UsdPrefPayloadMenuName, priority = 0)]
+    static void LoadAllPayloads()
+    {
+        var enabled = Menu.GetChecked(UsdPrefPayloadMenuName);
+        EditorPrefs.SetBool(PayloadPolicyPrefName, !enabled);
+        Menu.SetChecked(UsdPrefPayloadMenuName, !enabled);
+    }
+
+    [MenuItem(UsdPrefPayloadMenuName, true)]
+    static bool LoadAllPayloadsValidate() {
+        var enabled = EditorPrefs.GetBool(PayloadPolicyPrefName, true);
+        Menu.SetChecked(UsdPrefPayloadMenuName, enabled);
+        return true;
+    }
+
+    [MenuItem(UsdPrefMaterialMenuName, priority = 0)]
+    static void LoadAllPreviewSurface()
+    {
+        var enabled = Menu.GetChecked(UsdPrefMaterialMenuName);
+        EditorPrefs.SetBool(MaterialPolicyPrefName, !enabled);
+        Menu.SetChecked(UsdPrefMaterialMenuName, !enabled);
+    }
+
+    [MenuItem(UsdPrefMaterialMenuName, true)]
+    static bool LoadAllPreviewSurfaceValidate() {
+        var enabled = EditorPrefs.GetBool(MaterialPolicyPrefName, false);
+        Menu.SetChecked(UsdPrefMaterialMenuName, enabled);
+        return true;
+    }
+
+    [MenuItem("USD/Import as GameObjects", priority = 20)]
     public static void MenuImportAsGameObjects() {
       var scene = InitForOpen();
       if (scene == null) {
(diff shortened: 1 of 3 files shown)


package/com.unity.formats.usd/Editor/Scripts/Behaviors/UsdMenu.cs
```diff
@@ -19,90 +19,118 @@
...
-    static bool EnableMenuExportSelectedWithChildren() {
-      return Selection.gameObjects.Length > 0;
-    }
-    [MenuItem("USD/Export Selected with Children", priority = 50)]
-    static void MenuExportSelectedWithChildren() {
-      ExportSelected(BasisTransformation.SlowAndSafe);
-    }
+        public static Scene InitForOpen()
+        {
+            string path = EditorUtility.OpenFilePanel("Import USD File", "", "usd,usda,usdc,abc");
```
clusty: I suggest strongly to not change massively the indentation during a feature PR: close to impossible to see the actual changes.

package/com.unity.formats.usd/Editor/Scripts/Behaviors/UsdMenu.cs
```diff
@@ -148,267 +176,365 @@ public class UsdMenu : MonoBehaviour {
...
-      string path = scene.FilePath;
-
-      var invalidChars = Path.GetInvalidFileNameChars();
-      var prefabName = string.Join("_", GetPrefabName(path).Split(invalidChars,
-          System.StringSplitOptions.RemoveEmptyEntries)).TrimEnd('.');
+        [MenuItem(UsdPrefPayloadMenuName, priority = 0)]
+        static void LoadAllPayloads()
+        {
+            var enabled = Menu.GetChecked(UsdPrefPayloadMenuName);
+            EditorPrefs.SetBool(PayloadPolicyPrefName, !enabled);
```
clusty: I suggest you "steal" my SavedBool class altogether: that allows you to create it with your funky ID, and from that point onwards you treat as a regular bool. also that class can store the bool data. the menu just adds the tick according to this value.

clusty (approved):

package/com.unity.formats.usd/Editor/Scripts/Behaviors/UsdMenu.cs
jcowles: +1 -- if you agree, I'm actually going to hold off on reviewing until after that

jcowles:
There are actually quite a lot of potential "pipeline preferences", which makes me think that perhaps we should have a configuration asset, like HDRP...

package/com.unity.formats.usd/Editor/Scripts/Behaviors/UsdMenu.cs
judubu: Isn't it what the commit selection menu is for? Or at least it's how I used it before. 
I can get rid of the reformat in this PR and just do a pass on the whole package instead. Would you prefer that?

jcowles: Unless the reformatting is really minor, I usually break it into its own PR

But the second issue is that this is also reformating this one bit of code with a different standard from the rest of the repository.

How about this: I think this PR is fine as-is (it looks like the reformatting is in its own commit already), but can you do a follow up PR that reformats the rest of the repo to Unity's formatting conventions?

judubu: Removed the formatting commit and pushed a new formatting PR #229

package/com.unity.formats.usd/Editor/Scripts/Behaviors/UsdMenu.cs
judubu: Added in #9b02130

package/com.unity.formats.usd/Editor/Scripts/Behaviors/UsdMenu.cs
```diff
@@ -12,16 +12,27 @@
...
 namespace Unity.Formats.USD {
- 
-  public class UsdMenu : MonoBehaviour {
+
+  public static class UsdMenu
+  {
+    const string PayloadPolicyPrefName = "USD.LoadAllPayloads";
+    const string MaterialPolicyPrefName = "USD.LoadPreviewSurface";
+
+    const string UsdPrefPayloadMenuName = "USD/Preferences/Load all payloads";
```
cguer: It looks like most menu items are [title case](https://en.wikipedia.org/wiki/Title_case). For consistency -> `Load All Payloads`

package/com.unity.formats.usd/Editor/Scripts/Behaviors/UsdMenu.cs
```diff
@@ -12,16 +12,27 @@
...
+    const string UsdPrefMaterialMenuName = "USD/Preferences/Load preview surface";
```
cguer: Capitalization -> `Load Preview Surface`

package/com.unity.formats.usd/Editor/Scripts/Behaviors/UsdMenu.cs
cguer: Would it make sense to add `on Import` to the menu labels? Like `Load All Payloads on Import`

In (rare-ish) cases, a user might wrongly assume that toggling these preferences will also update existing USD assets in the scene.

cguer (changes requested):
I get the following error when installing the package. It is preventing me from further testing:
```
Library\PackageCache\com.unity.formats.usd@b0db90b49447-1607034678910\Editor\Scripts\Behaviors\UsdMenu.cs(34,12): error CS0122: 'SavedBool' is inaccessible due to its protection level
```

----

As more general comment, do we know if there is a convention for the location of package settings? Menu toggles vs. Preferences window vs. Project Settings Window

I don't yet have a strong opinion either way but I'm wondering if it's something that users would like to set on a per-project basis 🤔

thomas-tu:
> As more general comment, do we know if there is a convention for the location of package settings? Menu toggles vs. Preferences window vs. Project Settings Window
> 
> I don't yet have a strong opinion either way but I'm wondering if it's something that users would like to set on a per-project basis 🤔

* Editor Preferences -> Anything that is shared between projects on your machine. Useful for general+user configurations.
* Project Preferences -> Anything that apply **only** to your project. That's why to have so many settings here, they are all defined per project.
* Menu toggles -> I don't think that is standard. They pollute menus and go against the usage of Preferences or Project Settings.

Every time a new setting is exposed, we need to define its scope. Is it define per project? per user?

It matters to stay consistent so users don't get lost. @JulienDidierjean can probably confirm what is the best way.

package/com.unity.formats.usd/Editor/Scripts/Behaviors/UsdMenu.cs
judubu: I'd prefer to keep this one like that for now and work on a improved way of settings import prefs, such as an UsdPrefs asset like Jeremy suggested.

judubu:
> It matters to stay consistent so users don't get lost. @JulienDidierjean can probably confirm what is the best way.

I already discussed the matter with @JulienDidierjean and even if not standard he was ok with it. Recorder is using the same trick as well. 
@thomas-tu would you add a usd section to the Preferences even for those 2 options? I kind of like the idea of a hrdp style config asset though.

judubu:
Closing the PR for rework. Planned for 2.1.0

PR #234, opened 2021-01-19
CI: Allow true isolation

mfe:
## Purpose of this PR

Added the option to test the isolation + 2 missing dependencies.
2019.4 and 2020.2 are still failing because of Timeline dependencies not yet adapted to true isolation.
Will ask what is the timeframe to fix previous versions.

We could also update our timeline dependency (trunk and 2021.1 are using 1.5.1-pre.3).
I double checked, CI is back to green when doing so.

While I was there I updated the Unity test versions.

**Complexity:**
Minimal 

**Halo Effect:**
Minimal

### DIFF
--- a/.yamato/upm-ci.yml
+++ b/.yamato/upm-ci.yml
@@ -1,6 +1,7 @@
 test_editors:
   - version: trunk
-  - version: 2020.1
+  - version: 2020.2
+  - version: 2021.1
   - version: 2019.4
 test_platforms:
   - name: win
--- a/.yamato/upm-ci.yml
+++ b/.yamato/upm-ci.yml
@@ -37,7 +37,7 @@ test_{{ platform.name }}_{{ editor.version }}:
   commands:
       - npm install upm-ci-utils@latest -g --registry https://artifactory.prd.cds.internal.unity3d.com/artifactory/api/npm/upm-npm
       - upm-ci package test --unity-version {{ editor.version }} --package-path package/com.unity.formats.usd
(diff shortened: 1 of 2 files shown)


.yamato/upm-ci.yml
```diff
@@ -1,6 +1,7 @@
 test_editors:
   - version: trunk
-  - version: 2020.1
+  - version: 2020.2
+  - version: 2021.1
   - version: 2019.4
```
clusty, line 5: would also add trunk

clusty, line 5: would also add trunk

package/com.unity.formats.usd/package.json
```diff
@@ -1,6 +1,8 @@
 {
     "dependencies": {
-        "com.unity.timeline" : "1.0.0"
+        "com.unity.timeline" : "1.0.0",
+        "com.unity.modules.screencapture": "1.0.0",
+        "com.unity.modules.physics": "1.0.0"
```
clusty, line 5: for my curiosity, what part of physics is USD using ?

clusty (approved):

.yamato/upm-ci.yml
mfe: There is already trunk

package/com.unity.formats.usd/package.json
mfe: RigidBody

mfe:
Timeline dependencies are expected to be fixed for < 2021 soon.
So waiting for this to happen before merging so our CI remains green.

judubu (approved):

PR #235, opened 2021-01-20
CMake based build on all platforms (USDU-115)

judubu:
## Purpose of this PR

**Ticket/Jira #:** https://jira.unity3d.com/browse/USDU-115

This PR introduces a multi-platform CMake based build to make the bindings easier to build on all 3 platforms and to align with the USD build system and other common build system in the VFX/Animation industry.

It currently works on Windows but has some issues on Linux and OSX.

Check the updated BUILDING.md for requirements and detailed build steps.

What changed:
* new multi-platform bin/build.py entry point
* download of usd binaries from stevedore if you're on Unity network (could potentially be made public)
* all the various manual steps have been automated through python or cmake (type info generation, swig callback decoration for IL2CPP, runtime dependencies install, usd plugins install)
* 2 different components UsdCs and USD.NET
* yamato jobs to build UsdCs

What's not working yet @cpredmann @jcowles :
* linux build: worked at some point but now the target dependencies in UsdCs are messed up (tries to generate the SWIG bindings before running the type info gen)
* osx build: the osx image I'm using is currently broken, never tried to generate a bundle with cmake yet
* rpath might need some more work


BUILDING.md
```diff
@@ -21,89 +19,72 @@ Start by understanding the layout of the source code:
...
- * The Python step has been tested with Ptyhon 2.7 and requires the USD
-   Python files to be importable via PYTHONPATH.
- * The SWIG step requires the SWIG 3.0.12 executable on your system PATH.
+In a terminal, running the following command from the root of the repository will build UsdCs by default:
 
-By setting USD_LOCATION_PYTHON to the root install directory of a USD
-python build, all scripts will work without setting the system PATH or
-PYTHONPATH. The variable USD_LOCATION should be set to a USD build without
-python, to minimize runtime dependencies of the final plugin.
+`python3 bin/build.py 20.08 $USD_LOCATION 2019.4`
```
clusty, line 65: would be nice if you could use USD_LOCATION only if you can't download or are an advanced user:
Basically:
If !USD_LOCATION set
    If Can Download
       Download and set USD_LOCATION
    Warn user to set USD_LOCATION

use USD_LOCATION

:)

BUILDING.md
```diff
@@ -21,89 +19,72 @@ Start by understanding the layout of the source code:
...
-### OSX / Darwin
+We typically use the standard build steps from the USD repo instructions. The exact command line we are using are:
+* with python support: `python3 build_scripts/build_usd.py --build-monolithic --alembic --no-imaging --no-examples --no-tutorials ../artifacts/usd-v20.08`
+* no python support:   `python3 build_scripts/build_usd.py --build-monolithic --alembic --no-python --no-imaging --no-examples --no-tutorials ../artifacts/usd-v20.08_no_python`
 
-Similarly for OSX, an XCode project is provided in the src directory, called
-UsdCs.xcodeproj. The OSX build will produce a .bundle file, into which all
-dependent dylibs must be manually copied, as well as the USD plugInfo.json
-files. See the existing bundle as an example of the correct structure.
+**Unity developers**: USD binaries are built with each new release and pushed on Stevedore. See next section to download them automatically when building the bindings.
```
clusty, line 51: Would be nice if these assets could be posted to the public stevedore so all external users could benefit from this goodness.

clusty:
Since the binding builds are not managed automatically for all platforms, You need to add a C# test to make sure no platform was left behind :) eg: version the bindings and add a C# unit test that asserts that they are the correct version on all platforms.

clusty (approved):

BUILDING.md
judubu: Fixed in e38eb675
Now downloads in ./artifacts . No need for a USD_LOCATION var anymore

BUILDING.md
judubu: I'll add that in the backlog. https://jira.unity3d.com/browse/USDU-156

judubu:
@clusty Added a task to version the bindings and test against all supported platforms: https://jira.unity3d.com/browse/USDU-157

PR #240, opened 2021-04-08
Move unity api to pkg (USDU-137)

judubu:
## Purpose of this PR

**Ticket/Jira #:** [USDU_137](https://jira.unity3d.com/browse/USDU-137)

This PR moves the Unity specific part of the bindings (USD.NET.Unity) inside the package to allow faster iterations and better test environment. 

The old copyright notices have been removed from the files and the unit/regression tests have been moved from the VS project to the package. Assembly defs have been updated accordingly.

## Testing

Tests from the old VS project have been moved and updated to the Unity testing framework.

No impact on perfs

## Overall Product Risks

**Complexity:**

Low

**Halo Effect:**

Low

## Additional information

**Note to reviewers:**

**Reminder:**


package/com.unity.formats.usd/Tests/USD.NET.Unity/InstancingTests.cs
```diff
@@ -0,0 +1,97 @@
...
+using pxr;
+using UnityEngine;
+
+namespace USD.NET.Unity.Tests
+{
+    class InstancingTests : UsdTests
+    {
+        Scene CreatePointInstancerScene()
+        {
+            var tmpScenePath = CreateTmpUsdFile("PointInstancerTest.usda");
```
clusty, line 13: I would make the file name be random, just to make sure you never fall on the same file twice.

package/com.unity.formats.usd/Tests/USD.NET.Unity/UsdNetUnityTests.cs
```diff
@@ -0,0 +1,159 @@
...
+            m_filesToDelete.Add(filePath);
+            return filePath;
+        }
+
+        [TearDown]
+        public void DeleteTmpFiles()
+        {
+            foreach (var file in m_filesToDelete)
+            {
+                File.Delete(file);
```
clusty: You Might want to delete the meta file as well if it exists. Otherwise you'll get a one time warning telling you it found a meta without the file

clusty: Also you might delete the file from the "delete list" once it's deleted cause this function could be called multiple times in theory.

clusty (approved):

PR #247, opened 2021-05-24
Blend Shape Importing

cpredmann:
This PR enables importing of blend shapes from USD
- Toggle for enabling or disabling support
- Blend shapes are supported by reading from BlendShapeSample
- Animated blend shapes are supported from SkelAnimSample
- Skinned mesh renderers with or without bones are supported


phocodes (approved):

jcowles:
What's up with this PR? @judubu @cpredmann

cpredmann:
> What's up with this PR? @judubu @cpredmann

I don't think it's being pulled in, so should likely just close it

PR #249, opened 2021-06-01
Move usdnet to package (#161)

judubu:
## Purpose of this PR

**Ticket/Jira #:**https://jira.unity3d.com/browse/USDU-161

This moves the USD.NET API to the package to make it easier to iterate on the bindings and simplify the build pipeline. There's a small cost in compile time in most cases and with the biggest regression on IL2CPP player builds (12s to 1m16s, initial build only). We still think that the benefits overweight the regression.

Building UsdCs now installs the generated bindings in the USD.NET directory in the package.
USD.NET can still be built with CMake for the folks wanting to just use the bindings but not the usd package.

Bringing the API inside the package exposed compile warnings and the bindings have been updated to fix those.

## Testing

Old tests from the Visual Studio solution have been moved and updated to work in USD. They haven't been modified further than making sure they work in Unity Test Runner.

Compile time performance benchmarks
                                     OLD         NEW (USD.NET in package)
package load             17.65       19.54
reimport                      6.75        11.29
recompile                    4.89        4.54
recompile USD.NET                    6.61
il2cpp build 1            12.38       1'16.0
il2cpp build 2            10.64       11.45
mono build 1              7.06        9.25
mono build 2              6.82        9.05

## Overall Product Risks

**Complexity:**

Low

**Halo Effect:**

Medium

## Additional information

**Note to reviewers:**

**Reminder:**


package/com.unity.formats.usd/Tests/USD.NET.Unity/XformTests.cs
```diff
@@ -0,0 +1,64 @@
+﻿using NUnit.Framework;
+using UnityEngine;
+
+namespace USD.NET.Unity.Tests
+{
+    class XformTests : UsdTests
+    {
+        [Test]
+        public static void XformTest()
```
clusty, line 9: static ?

clusty (approved):

PR #253, opened 2021-06-10
Abort build post-process for non desktop platforms

judubu:
## Purpose of this PR

Build post process was failing for non desktop builds.
Now aborting the post process when building for non supported platforms.


package/com.unity.formats.usd/Editor/Scripts/UsdBuildPostProcess.cs
```diff
@@ -39,6 +39,10 @@ public static void OnPostprocessBuild(BuildTarget target, string pathToBuiltProj
             {
                 destination = pathToBuiltProject.Replace(".exe", "_Data/Plugins");
             }
+            else
```
clusty, line 42: I would throw an exception or a Debug.LogError at least.
You don't want silent errors on crazy platforms.

judubu: Is it really an issue to build a project on mobile platform with the usd package, even if not used?

clusty: ppl should know that they USD was not used for the iOS build.
I think I agree with you that stopping the build might be intense, but the very least there should be message about the fact that device builds are not supported.

judubu: Good point, I'll add an extra message

judubu: I added a warning message on non desktop builds.

clusty (approved):

PR #260, opened 2021-07-31
Add type alias support to the reverse bindings (#USDU-3 part 2)

judubu:
## Purpose of this PR

**Ticket/Jira #:**

USD allows certain attribute type to be given _[roles](https://graphics.pixar.com/usd/docs/api/_usd__page__datatypes.html#Usd_Roles)_ (aliases) to help unambiguously determine how data should be interpreted.
Ex: 
- float2 can be given the texcoord2 when used for a uv primvar
- float3 can be given the point/normal/vector role ...

This is a problem when using our reverse bindings which let the USD type drive the C# type the primvar data should deserialized to.  The reverse bindings are useful in the case a given attribute can be defined as multiple types. Typically UVs primvars can be defined as float2/float3/half2/half3/double2/double3 and as such can be also given the role of texcoord2f/texcoord3f/texcoord2h/texcoord3h/texcoord2d/texcoord3d .... As we already have a 1-to-1 mapping to Unity type for the standard types (float2, float3, ...) we can't augment the mapping to support roles.

This PR adds a way to define aliases for a given type and allow the reverse binding to convert roles to their canonic type before doing the type lookup.

## Testing

## Overall Product Risks

**Complexity:**

**Halo Effect:**

## Additional information

**Note to reviewers:**

**Reminder:**

### DIFF
--- a/package/com.unity.formats.usd/Dependencies/USD.NET.Unity/UnityTypeBindings.cs
+++ b/package/com.unity.formats.usd/Dependencies/USD.NET.Unity/UnityTypeBindings.cs
@@ -55,12 +55,12 @@ namespace USD.NET.Unity
                 (object obj) => UnityTypeConverter.Vector2ToVec2f((Vector2)obj),
                 (pxr.VtValue value) => UnityTypeConverter.Vec2fToVector2(value),
                 SdfValueTypeNames.Float2));
-            binder.typeAliases.Add(SdfValueTypeNames.TexCoord2f.GetAsToken(), SdfValueTypeNames.Float2.GetAsToken());
+            binder.AddTypeAlias(SdfValueTypeNames.TexCoord2f, SdfValueTypeNames.Float2);
             binder.BindType(typeof(Vector3), new UsdTypeBinding(
                 (object obj) => UnityTypeConverter.Vector3ToVec3f((Vector3)obj),
                 (pxr.VtValue value) => UnityTypeConverter.Vec3fToVector3(value),
                 SdfValueTypeNames.Float3));
-            binder.typeAliases.Add(SdfValueTypeNames.TexCoord3f.GetAsToken(), SdfValueTypeNames.Float3.GetAsToken());
+            binder.AddTypeAlias(SdfValueTypeNames.TexCoord3f, SdfValueTypeNames.Float3);
             binder.BindType(typeof(Vector4), new UsdTypeBinding(
                 (object obj) => UnityTypeConverter.Vector4ToVec4f((Vector4)obj),
                 (pxr.VtValue value) => UnityTypeConverter.Vec4fToVector4(value),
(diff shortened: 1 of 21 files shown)


package/com.unity.formats.usd/Dependencies/USD.NET/serialization/TypeBinder.cs
```diff
@@ -68,6 +69,12 @@ public class TypeBinder
         Dictionary<string, string> typeNameMapping = new Dictionary<string, string>();
         Dictionary<Type, Dictionary<pxr.TfToken, Enum>> enumMaps = new Dictionary<Type, Dictionary<pxr.TfToken, Enum>>();
 
+        /// <summary>
+        /// USD types can be aliased to represent different roles float2/texcoord2, vec3/point/color and drive the interpretation
+        /// Add aliases if your target application need to map multiple USD types to a single type.
+        /// </summary>
+        public Dictionary<string, string> typeAliases = new Dictionary<string, string>();
```
jcowles: Is there a reason this needs to be exposed publicly? If not, I think it would be better to add a wrapper API around it for adding aliases.

package/com.unity.formats.usd/Dependencies/USD.NET.Unity/UnityTypeBindings.cs
```diff
@@ -55,10 +55,12 @@ static public void RegisterTypes()
                 (object obj) => UnityTypeConverter.Vector2ToVec2f((Vector2)obj),
                 (pxr.VtValue value) => UnityTypeConverter.Vec2fToVector2(value),
                 SdfValueTypeNames.Float2));
+            binder.typeAliases.Add(SdfValueTypeNames.TexCoord2f.GetAsToken(), SdfValueTypeNames.Float2.GetAsToken());
```
jcowles: It's odd that this API requires `GetAsToken`, can we do that internally, maybe via a wrapper API?

jcowles (commented):
This is awesome!

I do think we should keep the data private (typeAliases) and wrap it with a public API, but other than that, it looks great.

clusty (approved):
Aggree with @jcowles 's comments

package/com.unity.formats.usd/Dependencies/USD.NET/serialization/TypeBinder.cs
judubu: Not really ... I'll add a wrapper as advised.

judubu: Done in https://github.com/Unity-Technologies/usd-unity-sdk/pull/260/commits/4e5e928f3618110e5ec30616b01f6e413329a25c

package/com.unity.formats.usd/Dependencies/USD.NET.Unity/UnityTypeBindings.cs
judubu: Done in https://github.com/Unity-Technologies/usd-unity-sdk/pull/260/commits/4e5e928f3618110e5ec30616b01f6e413329a25c

jcowles (approved):
Ship it!

PR #261, opened 2021-07-31
MeshImporter/XFormImporter refactor (USDU-3 part 3)

judubu:
## Purpose of this PR

**Ticket/Jira #:**
https://jira.unity3d.com/browse/USDU-3
https://jira.unity3d.com/browse/USDU-10

USD geometric Primvars can be given different kind of [interpolation ](https://graphics.pixar.com/usd/docs/api/class_usd_geom_primvar.html#Usd_InterpolationVals) across the uv space of a surface primitive:
- **constant**: one value per-prim
- **uniform**: one value per-face
- **vertex/varying**: one value per-vertex
- **facevarying**: one value per-vertex per-face

Each primvar can also be individually indexed for data-compression purpose.

This is an issue for Unity which requires all attributes of a mesh to be either interpolated per vertex or per-vertex per-face. 
As a consequence it forces us to sanitize the data on the way in before being pushed to GameObjects.

The sanitization was previously done in the MeshImporter right before setting the Unity mesh buffers which was single threaded and made it impossible to properly cross check all attributes arrays for their interpolation value.

This PR introduces a way to perform sanitization of the Mesh and XForm data right after deserialization while using the job system to lessen the impact of interpolation conversion. For that a new ISanitizable interface and 2 new classes SanitizedMeshSample and XFormMeshSample have been added and the jobified ReadAllJob struct have been modified to provide sanitization service after deserialization. 

The sanitization code takes care of:
    - xform conversion
    - triangulation
    - de-indexation of attributes
    - attribute scope conversion

As a consequence the MeshImporter has been greatly simplified.

The SkeletonImporter has been changed and the JointWeights and JointIndices are now deserialized/loaded by the MeshImporter to avoid aan extra mesh deserialization/sanitization.

## Testing

The sanitization service has almost 100% coverage but not the MeshImporter itself.

@cguer You should check the import with all the regular suspect (kitchen, human female, [walking teapot](https://pixar-community-production.s3.us-west-1.amazonaws.com/Siggraph/2020/AR_Walking_Teapot_2020_Straight_Line.usdz) ) and do a before after. 

I quickly profile the changes and they are a little bit slower both for static and animated data. Although GC has been greatly reduced. I'll update with the actual numbers after my PTO.

## Overall Product Risks

**Complexity:**

High

**Halo Effect:**

High

## Additional information

**Note to reviewers:**

Check the SanitizedMeshSample for the actual sanitization code and the MeshImporter to see how the sanitized data is pushed to Unity Meshes.
There are some minor changes to the deserialization process to flag colors as actual primvars which give us access to the indices and interpolation values.
Check the test usda file to get a feel for the kind of crappy edge case that can come up and if I forgot anything.

**Reminder:**

### DIFF
--- a/package/com.unity.formats.usd/Dependencies/USD.NET.Unity/Geometry/MeshSampleBase.cs
+++ b/package/com.unity.formats.usd/Dependencies/USD.NET.Unity/Geometry/MeshSampleBase.cs
@@ -24,8 +24,7 @@ namespace USD.NET.Unity
         public int[] faceVertexIndices;
         public Vector3[] points;
         public Vector3[] normals;
-        [VertexData]
-        public Vector4[] tangents;
+        [VertexData] public Vector4[] tangents;
 
         // Regarding UVs: this feels like a very specific solution for "default primvar data", which
         // is fine, but this type of data may be specific to a given pipeline, though here it is
@@ -39,19 +38,10 @@ namespace USD.NET.Unity
         /// <remarks>
         /// UV object types should be Vector{2,3,4}[], List of Vector{2,3,4}, or null.
         /// </remarks>
-        [VertexData] public object st;
-
-        /// <summary>
-        /// When primvars:st:indices are specified, the st texture coordinates are indexed like
-        /// vertex positions.
-        /// </summary>
-        [UsdNamespace("primvars:st")]
-        public int[] indices;
-
-        // These are Unity friendly UV sets.
-        [VertexData] public object uv;
-        [VertexData] public object uv2;
-        [VertexData] public object uv3;
-        [VertexData] public object uv4;
+        public Primvar<object> st = new Primvar<object>();
+        public Primvar<object> uv = new Primvar<object>();
+        public Primvar<object> uv2 = new Primvar<object>();
+        public Primvar<object> uv3 = new Primvar<object>();
+        public Primvar<object> uv4 = new Primvar<object>();
     }
 }
--- a/package/com.unity.formats.usd/Dependencies/USD.NET.Unity/Geometry/MeshSampleBase.cs
+++ b/package/com.unity.formats.usd/Dependencies/USD.NET.Unity/Geometry/MeshSampleBase.cs
@@ -36,7 +36,7 @@ namespace USD.NET.Unity
         /// When not explicitly specified by the shader, "st" should be considered the default uv set.
         /// </summary>
         /// <remarks>
-        /// UV object types should be Vector{2,3,4}[], List of Vector{2,3,4}, or null.
+        /// UV object types should be Vector{2,3}[], List of Vector{2,3}, or null.
         /// </remarks>
         public Primvar<object> st = new Primvar<object>();
         public Primvar<object> uv = new Primvar<object>();
--- a/package/com.unity.formats.usd/Dependencies/USD.NET/serialization/Primvar.cs
+++ b/package/com.unity.formats.usd/Dependencies/USD.NET/serialization/Primvar.cs
@@ -142,6 +142,10 @@ namespace USD.NET
             return typeof(T);
         }
 
+        public bool IsArray => typeof(T).IsArray;
+
+        public int Length => IsArray && value != null ? (value as Array).Length : 0;
+
         // Furture work: support IdTargets. See "ID Attribute API" here:
         // http://graphics.pixar.com/usd/docs/api/class_usd_geom_primvar.html
     }
--- a/package/com.unity.formats.usd/Dependencies/USD.NET/serialization/Primvar.cs
+++ b/package/com.unity.formats.usd/Dependencies/USD.NET/serialization/Primvar.cs
@@ -144,7 +144,7 @@ namespace USD.NET
 
         public bool IsArray => typeof(T).IsArray;
 
-        public int Length => IsArray && value != null ? (value as Array).Length : 0;
+        public int Length => (IsArray && value != null) ? (value as Array).Length : 0;
 
         // Furture work: support IdTargets. See "ID Attribute API" here:
         // http://graphics.pixar.com/usd/docs/api/class_usd_geom_primvar.html
--- a/package/com.unity.formats.usd/Runtime/Scripts/IO/Geometry/CubeImporter.cs
+++ b/package/com.unity.formats.usd/Runtime/Scripts/IO/Geometry/CubeImporter.cs
@@ -79,7 +79,6 @@ namespace Unity.Formats.USD
                 {
                     // Constant color can just be set on the material.
                     mat = options.materialMap.InstantiateSolidColor(usdCube.colors.value[0].gamma);
-                    Debug.Log("constant colors assigned");
                 }
                 else if (usdCube.colors.Length == 6)
                 {
--- a/package/com.unity.formats.usd/Runtime/Scripts/IO/Geometry/CubeImporter.cs
+++ b/package/com.unity.formats.usd/Runtime/Scripts/IO/Geometry/CubeImporter.cs
@@ -78,7 +78,7 @@ namespace Unity.Formats.USD
                 if (usdCube.colors.Length == 1)
                 {
                     // Constant color can just be set on the material.
-                    mat = options.materialMap.InstantiateSolidColor(usdCube.colors[0].gamma);
+                    mat = options.materialMap.InstantiateSolidColor(usdCube.colors.value[0].gamma);
                     Debug.Log("constant colors assigned");
                 }
                 else if (usdCube.colors.Length == 6)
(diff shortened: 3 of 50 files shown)


clusty (approved):
Only comment was around need for both: MeshSample and SanitizedMeshSample & friends classes.

package/com.unity.formats.usd/Dependencies/USD.NET.Unity/Geometry/MeshSampleBase.cs
```diff
@@ -37,21 +36,12 @@ public class MeshSampleBase : PointBasedSample
         /// When not explicitly specified by the shader, "st" should be considered the default uv set.
         /// </summary>
         /// <remarks>
-        /// UV object types should be Vector{2,3,4}[], List of Vector{2,3,4}, or null.
+        /// UV object types should be Vector{2,3}[], List of Vector{2,3}, or null.
```
jcowles, line 39: Why just 2,3 components?

package/com.unity.formats.usd/Dependencies/USD.NET/serialization/Primvar.cs
```diff
@@ -142,6 +142,10 @@ public Type GetValueType()
             return typeof(T);
         }
 
+        public bool IsArray => typeof(T).IsArray;
+
+        public int Length => IsArray && value != null ? (value as Array).Length : 0;
```
jcowles: Nit pick, but I would add parens around (IsArray && value != null) for readability.

package/com.unity.formats.usd/Runtime/Scripts/IO/Geometry/CubeImporter.cs
```diff
@@ -78,7 +78,7 @@ public static class CubeImporter
                 if (usdCube.colors.Length == 1)
                 {
                     // Constant color can just be set on the material.
-                    mat = options.materialMap.InstantiateSolidColor(usdCube.colors[0].gamma);
+                    mat = options.materialMap.InstantiateSolidColor(usdCube.colors.value[0].gamma);
                     Debug.Log("constant colors assigned");
```
jcowles: I think we should remove this debug.log

package/com.unity.formats.usd/Runtime/Scripts/IO/Geometry/MeshImporter.cs
```diff
@@ -464,44 +483,39 @@ public static GeometrySubsets ReadGeomSubsets(Scene scene, string path)
...
                 {
+                    var usdIndices = usdMesh.faceVertexIndices;
                     unityMesh.subMeshCount = geomSubsets.Subsets.Count;
-                    int subsetIndex = 0;
+                    var subsetIndex = 0;
                     foreach (var kvp in geomSubsets.Subsets)
                     {
-                        int[] faceIndices = kvp.Value;
-                        int[] triangleIndices = new int[faceIndices.Length * 3];
+                        var faceIndices = kvp.Value;
```
jcowles: Changing this to `var` makes the code harder to read, because we also don't know what types `kvp` holds.

package/com.unity.formats.usd/Runtime/Scripts/IO/SanitizedSamples.cs
```diff
@@ -0,0 +1,473 @@
...
+                        newValues.Add(values[next]);
+                    }
+
+                last += faceVertexCount[i];
+            }
+
+            values = newValues.ToArray();
+        }
+
+        void ConvertInterpolationToFaceVarying<T>(ref T[] values, int[] vertexIndices, bool changeHandedness = false, TfToken interpolation = null)
```
jcowles, line 372: I think it's worth adding a doc string for this function, since converting TO FaceVarying for import sounds strange.

I would expect things to be converted to Vertex/Varying interpolation for import.

package/com.unity.formats.usd/Runtime/Scripts/IO/SanitizedSamples.cs
```diff
@@ -0,0 +1,473 @@
...
+                Flatten(ref values, faceVertexIndices);
+            }
+            else if (interpolation == UsdGeomTokens.faceVarying) // 1 value per vertex per triangle
+            {
+                TriangulateAttributes(ref values, originalFaceVertexCounts, changeHandedness);
+            }
+        }
+
+        /// <summary>
+        /// Returns the interpolation of an data array based on the number of elements
```
jcowles, line 399: It's worth noting that this method, in general, will not always be correct.

It can be correct for constant and uniform interpolation, but varying/vertex/facevarying will sometimes be ambiguous.

package/com.unity.formats.usd/Runtime/Scripts/IO/SanitizedSamples.cs
```diff
@@ -0,0 +1,473 @@
...
+using System.Runtime.CompilerServices;
+using pxr;
+using USD.NET;
+using USD.NET.Unity;
+using UnityEngine;
+
+[assembly: InternalsVisibleToAttribute("Unity.Formats.USD.Tests")]
+namespace Unity.Formats.USD
+{
+    public interface ISanitizable
```
jcowles, line 12: I love this idea - it awesomely simplifies import code and isolates the transformation into concise operators

jcowles (approved):
Wow, this change is incredible :)

It's clear that a massive amount of effort went into this, really nice work!

I have a few minor comments, but overall it totally makes sense and is a great simplification.

cguer (approved):
QA approved 👍!

Tested on:
- 2021.2.0b10 (Windows + HDRP)

Sorry for the delay 🙈 ...

package/com.unity.formats.usd/Dependencies/USD.NET.Unity/Geometry/MeshSampleBase.cs
judubu: Because I could not find any reference to TexCoord4f. What are UV coords of type float4 by the way?

package/com.unity.formats.usd/Runtime/Scripts/IO/SanitizedSamples.cs
judubu: You're right, indexed arrays could easily mess it up and they are probably frequentely used.  
Indexed Vertex/Varying arrays will be shorter that points.length and indexed FaceVarying arrays will be shorter than faceVertexIndices.length or even equal to points.length in a worst case scenario!
All those situation will cause the MeshImporter to fail creating a valid mesh.

judubu: @jcowles Now the issue to solve is that we don't deserialize the interpolation type for normals and tangents (we don't do the indices either). Although we do it for the colors and uvs by defining them as Primvars. Normals and tangents are not technically USD primvars but do you see an issue if I set them as Primvars?

package/com.unity.formats.usd/Runtime/Scripts/IO/SanitizedSamples.cs
judubu: Added in dd63d81

package/com.unity.formats.usd/Runtime/Scripts/IO/Geometry/MeshImporter.cs
judubu: Added in dd63d81

package/com.unity.formats.usd/Runtime/Scripts/IO/Geometry/CubeImporter.cs
judubu: Fixed in dd63d81

package/com.unity.formats.usd/Dependencies/USD.NET/serialization/Primvar.cs
judubu: Fixed in dd63d81

package/com.unity.formats.usd/Runtime/Scripts/IO/SanitizedSamples.cs
judubu: Rereading the doc I had some of that wrong:
- normals can be "builtin" or primvars
- only primvar normals can be indexed
- builtin normals are subject to interpolation but not indexing

Primvar normals should normally take precedence on builtin normals but we dont support them at the moment. It leaves us with non indexed builtin normals for which the GuessInterpolation method should work.

@jcowles  What is the ambiguous situation you had in mind?

package/com.unity.formats.usd/Dependencies/USD.NET.Unity/Geometry/MeshSampleBase.cs
jcowles: Unity has float4 uvs, which are useful sometimes for storing tangent data, etc

package/com.unity.formats.usd/Runtime/Scripts/IO/SanitizedSamples.cs
jcowles: Indexed primvars for sure, but there are degenerate non-indexed cases, where you have four faces and all faces share the same verts - uniform, vertex/varying, face varying are not distinguishable. Or more common cases, e.g. if you have a single face, then vertex & face-varying are not distinguishable.

I think what you have is fine, but it's worth calling this out.

Other primitives (such as curves) have even more aliasing.

judubu: Got it thanks for the update. Adding that as comment in the code.

PR #279, opened 2021-12-06
Fix Export Transform Overrides (#USDU-216)

judubu:
## Purpose of this PR

**Ticket/Jira #:** https://jira.unity3d.com/browse/USDU-216

This feature never worked, but there was an example that was exporting overrides properly. 
The use case is that you bring a usd asset in Unity, tweak a bunch of transforms to change the layout, animate a couple of transforms, and you want to export your changes in an override layer. This file should only contain what you changed, nothing else.

This PR fixes override export. You can call it from the USD menu or from use a new option in the USD Recorder window. 

usd_recorder_export_override

## Testing

New tests checking that the exported file only contains over prims for the modified transforms and contains no sublayers, 

And this is the result of exporting overrides for the kitchen after moving the stool:
```
#usda 1.0
(
    defaultPrim = "Kitchen_set"
    endTimeCode = 0
    startTimeCode = 0
    upAxis = "Z"
)

over "Kitchen_set"
{
    matrix4d xformOp:transform.timeSamples = {
        0: ( (1, 0, 0, 0), (0, 5.960464477539063e-8, 0.9999999403953552, 0), (0, -0.9999999403953552, 5.960464477539063e-8, 0), (0, 0, 0, 1) ),
    }
    uniform token[] xformOpOrder = ["xformOp:transform"]

    over "Props_grp"
    {
        over "North_grp"
        {
            over "SinkArea_grp"
            {
                over "SinkFloor_grp"
                {
                    over "StoolWooden_1"
                    {
                        matrix4d xformOp:transform.timeSamples = {
                            0: ( (1, 0, 0, 0), (0, 1, 0, 0), (0, 0, 1, 0), (-32.90289306640625, 47.29999923706055, 129.8000030517578, 1) ),
                        }
                        uniform token[] xformOpOrder = ["xformOp:transform"]
                    }
                }
            }
        }
    }
}

```

Note that exporting animated overrides from recorder is expensive as every xform attributes have to be checked against the original prims.

## Overall Product Risks

**Complexity:**

Minimal

**Halo Effect:**

Minimal

## Additional information

**Note to reviewers:**

Reproduction Steps:

- load an usd file
- move one of the imported object
- USD>Export Transform Overrides (usd usda)
- open the exported file, you get all prims exported as override

Desired behavior: you should only get the single xform prim you modified (and its parents) 

**Reminder:**


judubu:
@sebastienduverne on the doc review side I just need your approval for the exposed recorder attribute option and the changelog edit.

judubu:
I noticed the recorder option name appears truncated in the UI. Checking with Vlad offline.

package/com.unity.formats.usd/Runtime/Scripts/Behaviors/UsdAsset.cs
```diff
@@ -604,20 +605,18 @@ public void ExportOverrides(Scene sceneInWhichToStoreTransforms)
                     overs,
                     BasisTransformation.SlowAndSafe,
                     exportUnvarying: false,
-                    zeroRootTransform: true);
-
-                var rel = ImporterBase.MakeRelativePath(overs.FilePath, sceneToReference.usdFullPath);
-                GetFirstPrim(overs).GetReferences().AddReference(rel, GetFirstPrim(baseLayer).GetPath());
```
judubu, line 620: @jcowles Do you remember why you were adding a relative ref to the original prim? 
My view here is that until we provide tools to target changes to specific layers  I consider the override layer exported from Unity should not reference any other files. Any layering that needs to happen to use the overrides is out of scope and should be done by the user.

Note that I'm also removing the original layer l619

clusty (approved):

judubu:
Checked with Vlad, the layout issue is known in the recorder settings.
I shortened the following property labels by removing the Export prefix:
* Export Materials
* Export Transform Overrides

usd_recorder_export_override_label_fix

@sebastienduverne can you confirm it's allright

cguer (approved):
QA approved 👍!

Tested on:
- 2021.2.7f1 (Windows only)

----

My only comment is UX-related for the `Transform overrides` option in the Recorder. At first, this looks like an option to "include" additional data in the exported file but it totally changes what will be in the file. For example, enabling this option will not export _anything_ other than the modified values (a user might assume that all objects will still be there). Having a tooltip could help as well to clarify.

There isn't much space but ideally the option we be named something like `Export transform overrides only`. It could even be in a drop-down where users select if they want to export `everything` OR `overrides only`.

Anyways, this is not a blocker for me. Just the kind of comment I would make on a "real" package 😂

cguer (changes requested):
Oops, I missed something in my previous review. Recording with `Transform Overrides` enabled throws the following exception if the Source is _not_ the root gameobject of an existing USD:
```
NullReferenceException: Object reference not set to an instance of an object
UnityEditor.Formats.USD.Recorder.UsdRecorderInput.BeginRecording (UnityEditor.Recorder.RecordingSession session) (at Library/PackageCache/com.unity.formats.usd@3559bc2baa00-1642451023084/Editor/Recorder/UsdRecorderInput.cs:18)
UnityEditor.Recorder.Recorder.SignalInputsOfStage (UnityEditor.Recorder.ERecordingSessionStage stage, UnityEditor.Recorder.RecordingSession session) (at Library/PackageCache/com.unity.recorder@4.0.0-pre.2/Editor/Sources/Recorder.cs:299)
UnityEditor.Recorder.RecordingSession.BeginRecording () (at Library/PackageCache/com.unity.recorder@4.0.0-pre.2/Editor/Sources/RecordingSession.cs:152)
UnityEditor.EditorApplication:Internal_CallUpdateFunctions()
```

The error is logged if:
- The source is a "normal" gameobject
- The source is a child of a USD root

package/com.unity.formats.usd/Editor/Recorder/UsdRecorderInput.cs
```diff
@@ -11,6 +12,18 @@ public class UsdRecorderInput : RecorderInput
 
         protected override void BeginRecording(RecordingSession session)
         {
+            if (Context.exportTransformOverrides)
+            {
+                // Settings.
+                UsdAsset usdAsset = Settings.GameObject.GetComponentInParent<UsdAsset>(); // Get the UsdAsset component in this GameObject or its nearest parent
```
vickycl, line 18: @judubu Is this change ok with you?

vickycl (commented):
There's one test failing that I will investigate, but otherwise I've tested the repro and fixed the recorder issue, and I think it's in a good state :)

package/com.unity.formats.usd/Editor/Recorder/UsdRecorderInput.cs
judubu: Looks good. 👍

lee-aandrew:
Exporting as transform override only saves the changed object and its transform data

Using the Unity Recorder with USD Clip setting "Export X Form Overrides" only record the objects that has its transform changed (Tested both manually changing the transform of objects and running timeline animation)

Confirmed with Vicky's more recent fix the [error](https://github.com/Unity-Technologies/usd-unity-sdk/pull/279#pullrequestreview-854808058) reported by cguer has been fixed as well

lee-aandrew (approved):
~LGTM!~
NVM: https://github.com/Unity-Technologies/usd-unity-sdk/pull/279#pullrequestreview-1342089601

lee-aandrew (changes requested):
With these changes, the test case ExportSkeletalAnimation_ExportedPrimsHaveSkeletalTypes is failing

package/com.unity.formats.usd/Runtime/Scripts/Behaviors/UsdAsset.cs
vickycl: Hey @judubu, when exporting the Transform Overrides from the Recorder, we still keep the original scene reference, but we don't here. I don't feel strongly about whether or not we should include the relative ref, but I think they should probably be consistent. Is there some limitation that means for the recorder export we have to include this reference, or something else?

vickycl (approved):
Looks ready to land to me

cguer (approved):

package/com.unity.formats.usd/Editor/Recorder/UsdRecorderInput.cs
```diff
@@ -11,10 +12,28 @@ public class UsdRecorderInput : RecorderInput
...
+                    UnityEngine.Debug.LogError($"Unable to perform a 'transform overrides only' recording as <{Settings.GameObject.name}> is not a UsdAsset.");
+            }
             SceneExporter.SyncExportContext(Settings.GameObject, Context);
             SceneExporter.Export(Settings.GameObject,
                 Context,
                 zeroRootTransform: false);
+
+            if (Context.exportTransformOverrides)
+            {
+                // this is very brittle- if we have the chance of other sublayers in future we should store the index it was added at and only erase that one.
```
michaeljblain, line 34: Do we happen to have this logged as something we should improve?

package/com.unity.formats.usd/Runtime/Scripts/Behaviors/UsdAsset.cs
```diff
@@ -614,20 +615,20 @@ public void ExportOverrides(Scene sceneInWhichToStoreTransforms)
...
                 Debug.LogException(ex);
-                return;
             }
             finally
             {
                 if (overs != null)
                 {
+                    // Remove the reference to the original USD from the override file for flexibility in an asset pipeline
+                    // TODO: Make this an optional setting
+                    overs.Stage.GetRootLayer().GetSubLayerPaths().Erase(0);
```
michaeljblain, line 631: Similar to above comment. This may also be considered "brittle".

package/com.unity.formats.usd/Runtime/Scripts/IO/Geometry/XformExporter.cs
```diff
@@ -37,9 +35,32 @@ public static void ExportXform(ObjectContext objContext, ExportContext exportCon
...
+            if (exportContext.exportTransformOverrides)
+            {
+                UnityEngine.Profiling.Profiler.BeginSample("USD: Xform override check");
+                var sourceSample = new XformSample();
+                float tolerance = 0.0001f;
+                exportContext.scene.Read(path, sourceSample);
+                bool areClose = true;
+                for (int i = 0; i < 16; i++)
+                {
+                    if (Mathf.Abs(sample.transform[i] - sourceSample.transform[i]) > tolerance)
```
michaeljblain, line 50: Some concerns how this behaves for large numbers. Maybe we should be considering relative error in this. Is there a library function that can help with this? Maybe we could also add that case to the test.

michaeljblain (changes requested):
Looks good. Some minor comments.

package/com.unity.formats.usd/Editor/Recorder/UsdRecorderInput.cs
vickycl: Added: https://jira.unity3d.com/browse/USDU-362

package/com.unity.formats.usd/Runtime/Scripts/IO/Geometry/XformExporter.cs
vickycl: hmmm interesting point. Theoretically, I think the override should be exported irrelevant of relative scale. For example, if the whole Disney island were moved by half a metre I don't think that should be exported any different to if it was an individual character moving half a metre, even if the relative scales are different. Maybe I'm missing something in how Xforms interact in a hierarchy, though?

michaeljblain (approved):

lee-aandrew:
Tested again on Unity 2019.4.26f1
1. Imported Maneki.usd file into Unity
2. Changed "Coin" object by changing its transform values (position & rotation & scale)
3. Exported Transform override as .usda
4. Checked the USDA file only contains the transform changes
5. Manually added the "subLayers" reference to the .usda file
6. Imported the transform override file - the imported file contains the transform changes 👍 

----------------------------------

1. Imported Maneki.usd file into Unity
2. Set the Unity Recorder to be USD Clip & Export Transform Override Only
3. Changed "Coin" object by changing its transform values (position & rotation & scale) while the recorder was recording
4. Checked the resulting USDA file only contains the transform changes 👍 
5. Manually added the "subLayers" reference to the .usda file
6. Imported the .usda file into Unity as Timeline Clip
8. Applied the imported Timeline Clip to the original already imported Maneki object - the animation plays 👍 

Checked the test case that was previously failing "ExportSkeletalAnimation_ExportedPrimsHaveSkeletalTypes" is now passing 👍 
Confirmed the 3 new automated test cases are also passing 👍 

LGTM!

lee-aandrew (approved):

PR #295, opened 2022-02-24
Fix animated mesh properties

judubu:
## Purpose of this PR

**Ticket/Jira #:**

Sanitizing mesh data requires the original topology (face count and face indices) to be available at all times.
But to improve deserialization performance when reading animation data the static properties are filtered out and not available to the import logic.

A ConversionState interface has been added to hold Sample states and is stored inside the AccessMask initialized globally for all prims at the first frame of animation.

UnitTests to come

## Testing

## Overall Product Risks

**Complexity:**

**Halo Effect:**

## Additional information

**Note to reviewers:**

**Reminder:**


package/com.unity.formats.usd/Runtime/Scripts/IO/Geometry/MeshImporter.cs
```diff
@@ -55,6 +55,24 @@ public class MeshImportStrategy : IImporter
...
+            //     var hasPrimSource = primMap[meshPath].TryGetComponent(out primSource);
+            //     if (hasPrimSource)
+            //     {
+            //         contexts[i] = primSource;
+            //     }
+            //     else
+            //     {
+            //         contexts[i] = null;
+            //     }
+            // }
```
jcowles: Delete?

jcowles:
Seems like this is still not importing UVs correctly - is that a separate issue?

Test model:
https://sketchfab.com/3d-models/grass-variations-c84573687bf14f89938002df4ca0e696#download

3.0.0 FIX branch:
image
image


2.0.0
image
image


I hope to send review for PR this weekend

jcowles:
Discussing offline, I think the right thing to do is to first land this change and then land the UV fix after.

package/com.unity.formats.usd/Runtime/Scripts/IO/Geometry/MeshImporter.cs
```diff
@@ -535,7 +559,8 @@ public static void ImportSkinning(string path, SanitizedMeshSample usdMesh, Mesh
                 unityMesh.normals = usdMesh.normals;
                 Profiler.EndSample(); // Import Normals
             }
-            else if (ShouldCompute(options.meshOptions.tangents))
+            // If the mesh had normals have been sanitized don't
```
jcowles: This comment makes no sense to me :) maybe it's incomplete?

package/com.unity.formats.usd/Dependencies/USD.NET/serialization/AccessMask.cs
```diff
@@ -12,13 +12,43 @@
...
+    /// downstream logic needs some static properties they can be stored temporarily in a ConversionContext.
+    /// </remarks>
+    public interface IConversionState
+    {
+    }
+
+    /// <summary>
+    /// This class holds the data required by the deserializer to optimize reading of primitive with dynamic properties
+    /// </summary>
+    public class DeserializationContext : IEnumerable
```
jcowles: Would `StreamingContext` be a more accurate name?

package/com.unity.formats.usd/Runtime/Scripts/IO/SanitizedSamples.cs
```diff
@@ -409,15 +503,15 @@ internal TfToken GuessInterpolation(int count)
             {
                 return UsdGeomTokens.constant;
             }
-            if (count == originalFaceVertexCounts.Length)
+            if (IsTopologyBackedUp() && count == originalFaceVertexCounts.Length)
```
jcowles, line 554: Is there a scenario where GuessInterpolation /should/ be called, but the topology is NOT backed up?

If not, I think this should be an assert instead (and same a few lines down)

package/com.unity.formats.usd/Runtime/Scripts/IO/Scene/SceneImporter.cs
```diff
@@ -813,6 +813,7 @@ private static void RemoveComponent<T>(GameObject go)
                                 GameObject go = primMap[pathAndSample.path];
                                 NativeImporter.ImportObject(scene, go, scene.GetPrimAtPath(pathAndSample.path),
                                     importOptions);
+                                // TODO: should we restore the conversion context here?
```
jcowles: Comment still valid?

package/com.unity.formats.usd/Runtime/Scripts/IO/SanitizedSamples.cs
```diff
@@ -87,15 +175,18 @@ public void Sanitize(Scene scene, SceneImportOptions importOptions)
             if (changeHandedness)
                 ConvertTransform();
 
-            var santizePrimvars = importOptions.ShouldBindMaterials ||
-                scene.IsPopulatingAccessMask || scene.AccessMask != null;                        //this is true when reading from the timeline
+            var santizePrimvars = IsRestored() && arePrimvarsFaceVarying || // if the sample is restored we already know
+                importOptions.ShouldBindMaterials ||
+                scene.IsPopulatingAccessMask ||                                  // this is true when initializign prims from the timeline
```
jcowles: Comment typo

package/com.unity.formats.usd/Runtime/Scripts/IO/SanitizedSamples.cs
```diff
@@ -87,15 +175,18 @@ public void Sanitize(Scene scene, SceneImportOptions importOptions)
             if (changeHandedness)
                 ConvertTransform();
 
-            var santizePrimvars = importOptions.ShouldBindMaterials ||
-                scene.IsPopulatingAccessMask || scene.AccessMask != null;                        //this is true when reading from the timeline
+            var santizePrimvars = IsRestored() && arePrimvarsFaceVarying || // if the sample is restored we already know
```
jcowles: I would add parens here to make this easier to read - for example, is it `IsRestore && arePrimvarsFaceVarying`, or is it the entire list of conditions?

Also, would be good to complete the sentence in the comment - already know what? That the primvars need to be sanitized?

Actually, is it possible to just remove IsRestored entirely from the API? I feel booleans like this tend to complicate state and lead to bugs...

package/com.unity.formats.usd/Runtime/Scripts/IO/ReadJob.cs
```diff
@@ -96,7 +96,20 @@ public void Execute(int index)
...
-                sample.Sanitize(m_scene, m_importOptions);
+
+                var restorableSample = sample as IRestorable;
+                DeserializationContext deserializationContext = null;
+                m_scene.AccessMask?.Included.TryGetValue(m_paths[index], out deserializationContext);
+                if (restorableSample != null && deserializationContext != null)
+                {
+                    restorableSample.FromState(deserializationContext.state);
+                    sample.Sanitize(m_scene, m_importOptions);
+                    deserializationContext.state = restorableSample.ToState();
```
jcowles: I would add a comment here, explaining how the lifetime works, something like:

```
// If a context was populated, it means streaming playback is currently happening.
// In addition, if the sample is of type IRestorable, then some state must be stored to support efficient streaming.
// This state is saved immediately after sanitizing the data and downstream processes are not allowed to modify this state.
```

package/com.unity.formats.usd/Dependencies/USD.NET/serialization/AccessMask.cs
```diff
@@ -12,13 +12,43 @@
...
 namespace USD.NET
 {
+    /// <summary>
+    /// An interface to enable storage of the conversion state.
+    /// </summary>
+    /// <remarks>
+    /// In order to speed up the deserialization of animated prims we only load the dynamic properties. If the
+    /// downstream logic needs some static properties they can be stored temporarily in a ConversionContext.
+    /// </remarks>
+    public interface IConversionState
```
jcowles: Maybe `IPrimStreamingState`?

package/com.unity.formats.usd/Dependencies/USD.NET/serialization/AccessMask.cs
```diff
@@ -28,6 +58,6 @@ namespace USD.NET
     /// </remarks>
     public class AccessMask
     {
-        public Dictionary<SdfPath, HashSet<MemberInfo>> Included = new Dictionary<SdfPath, HashSet<MemberInfo>>();
+        public Dictionary<SdfPath, DeserializationContext> Included = new Dictionary<SdfPath, DeserializationContext>();
```
jcowles, line 58: This field is named `Included`, does the serialization context belong here, or as a sibling to this field?

`AccessMask` is also no longer a great name for this - it should be something more like, `PrimStreamingContextMap`, or something.

We could make these changes in a future PR, but maybe add a comment?

package/com.unity.formats.usd/Dependencies/USD.NET/USD.NET.api
```diff
@@ -14645,7 +14645,7 @@ namespace USD.NET
 {
     public class AccessMask
```
jcowles, line 14646: Out of curiosity, are these .api files auto-generated?

package/com.unity.formats.usd/Runtime/Scripts/IO/SanitizedSamples.cs
```diff
@@ -9,6 +9,31 @@
 [assembly: InternalsVisibleToAttribute("Unity.Formats.USD.Tests")]
 namespace Unity.Formats.USD
 {
+    /// <summary>
+    /// When a prim is sparsely animated only dynamic properties are read into the Sample.
+    /// The ConversionSate purpose is to hold the data (static properties or computed data) to ensure proper conversion.
+    /// </summary>
+    public interface IRestorable
```
jcowles, line 17: Do we need `IRestorable` and `IConversionState`?

package/com.unity.formats.usd/Runtime/Scripts/IO/SanitizedSamples.cs
```diff
@@ -17,6 +42,24 @@ public interface ISanitizable
         void Sanitize(Scene scene, SceneImportOptions importOptions);
     }
 
+    public class MeshConversionState : IConversionState
```
jcowles: If you rename `IConversionState` to `IPrimStreamingState`, I think this should also be `MeshStreamingState`

package/com.unity.formats.usd/Runtime/Scripts/IO/SanitizedSamples.cs
```diff
@@ -9,6 +9,31 @@
...
+    /// <summary>
+    /// When a prim is sparsely animated only dynamic properties are read into the Sample.
+    /// The ConversionSate purpose is to hold the data (static properties or computed data) to ensure proper conversion.
+    /// </summary>
+    public interface IRestorable
+    {
+        /// <summary>
+        /// Returns true if the internal data has been restored from State data
+        /// </summary>
+        public bool IsRestored();
```
jcowles: It would be awesome if we could remove this bool - for some reason, it triggers read flags for me, in terms of creating future bugs.

jcowles (approved):
I left a lot of comments, but they are mostly weak suggestions.

Overall, I think this is an awesome change - great work!

GeorgeIngUnity (approved):

package/com.unity.formats.usd/Dependencies/USD.NET/serialization/AccessMask.cs
```diff
@@ -12,13 +12,43 @@
...
 using pxr;
 
 namespace USD.NET
 {
+    /// <summary>
+    /// An interface to enable storage of the conversion state.
+    /// </summary>
+    /// <remarks>
+    /// In order to speed up the deserialization of animated prims we only load the dynamic properties. If the
+    /// downstream logic needs some static properties they can be stored temporarily in a ConversionContext.
```
vickycl: ~ConversionContext no longer exists- should this be a DeserializationContext?~

package/com.unity.formats.usd/Dependencies/USD.NET/serialization/AccessMask.cs
vickycl: I don't think calling this a state is very clear. A state implies it is the State of the Conversion, whereas it's more like a container for the data during the conversion. As we're storing the raw data deserialized from USD, maybe `IPrimDeserializedData`?

package/com.unity.formats.usd/Dependencies/USD.NET/serialization/AccessMask.cs
```diff
@@ -12,13 +12,43 @@
...
+    {
+    }
+
+    /// <summary>
+    /// This class holds the data required by the deserializer to optimize reading of primitive with dynamic properties
+    /// </summary>
+    public class DeserializationContext : IEnumerable
+    {
+        public HashSet<MemberInfo> dynamicMembers;
+        public IConversionState state;
```
vickycl: Could we make this more clear that it stores the staticMembers, perhaps by calling it `public IPrimDeserializedData staticMemberData` or some such?

package/com.unity.formats.usd/Dependencies/USD.NET/serialization/AccessMask.cs
vickycl: I agree AccessMask is vague and it would be good to have something clearer. I think Streaming implies something different though, unless I'm misunderstanding what is happening at import. Maybe `TimeVaryingDataMap`, something that calls out that it's specifically for data that is read over time. Then 'Included' makes a bit more sense as a name for the data that is time varying. Or even something more explicit- `TimeVaryingPrimToDeserializedDataMap`

package/com.unity.formats.usd/Runtime/Scripts/IO/Geometry/MeshImporter.cs
```diff
@@ -100,43 +118,43 @@ public void BeginReading(Scene scene, PrimMap primMap, SceneImportOptions import
...
+                            deserializationContext.dynamicMembers.Add(faceVertexIndices);
+                            deserializationContext.dynamicMembers.Add(orientation);
+                            deserializationContext.dynamicMembers.Add(points);
+                            deserializationContext.dynamicMembers.Add(normals);
+                            deserializationContext.dynamicMembers.Add(colors);
+                            deserializationContext.dynamicMembers.Add(st);
+                            deserializationContext.dynamicMembers.Add(uv);
+                            deserializationContext.dynamicMembers.Add(uv2);
+                            deserializationContext.dynamicMembers.Add(uv3);
+                            deserializationContext.dynamicMembers.Add(uv4);
```
vickycl, line 129: How come we need to add all of them if just one of the members is varying? Also could there be a better way of checking whether there's a varying one instead of searching the HashMap for each member?

I don't see anywhere where any of these types get added to the dynamicMembers outside of this function, so when will we hit this?

package/com.unity.formats.usd/Runtime/Scripts/IO/SanitizedSamples.cs
```diff
@@ -10,6 +10,31 @@
 [assembly: InternalsVisibleToAttribute("Unity.Formats.USD.Tests.Runtime")]
 namespace Unity.Formats.USD
 {
+    /// <summary>
+    /// When a prim is sparsely animated only dynamic properties are read into the Sample.
+    /// The ConversionSate purpose is to hold the data (static properties or computed data) to ensure proper conversion.
```
vickycl: Typo 'ConversionState'
Maybe expand in the comment specifically what `IRestorable` is for, ie copying static data into additional samples.

package/com.unity.formats.usd/Runtime/Scripts/IO/SanitizedSamples.cs
```diff
@@ -10,6 +10,31 @@
...
+    {
+        /// <summary>
+        /// Returns true if the internal data has been restored from State data
+        /// </summary>
+        bool IsRestored();
+
+        /// <summary>
+        /// Restore internal data from the State data
+        /// </summary>
+        void FromState(IConversionState state);
```
vickycl: If we rename the State rename these too

package/com.unity.formats.usd/Runtime/Scripts/IO/Geometry/MeshImporter.cs
```diff
@@ -457,7 +481,7 @@ public static void ImportSkinning(string path, SanitizedMeshSample usdMesh, Mesh
             //
             // Mesh Topology.
             //
-            if (options.meshOptions.topology == ImportMode.Import && usdMesh.faceVertexIndices != null)
```
vickycl, line 460: Should we check IsTopologyBackedUp() in here as we use originalFaceVertexCounts? I'm unsure whether this has introduced a route in which that could be null at this point?

package/com.unity.formats.usd/Runtime/Scripts/IO/ReadJob.cs
```diff
@@ -96,7 +96,22 @@ public void Execute(int index)
             if (ShouldReadPath(m_scene, m_paths[index]))
             {
                 m_scene.Read(m_paths[index], sample);
-                sample.Sanitize(m_scene, m_importOptions);
+
+                var restorableSample = sample as IRestorable;
```
vickycl, line 100: I think we should check this isn't null straight away, as there's no point doing the 'TryGetValue' if so

package/com.unity.formats.usd/Runtime/Scripts/IO/SanitizedSamples.cs
```diff
@@ -10,6 +10,31 @@
 [assembly: InternalsVisibleToAttribute("Unity.Formats.USD.Tests.Runtime")]
 namespace Unity.Formats.USD
 {
+    /// <summary>
+    /// When a prim is sparsely animated only dynamic properties are read into the Sample.
+    /// The ConversionSate purpose is to hold the data (static properties or computed data) to ensure proper conversion.
+    /// </summary>
+    public interface IRestorable
```
vickycl, line 17: If we keep it, I think if this is the only usecase for it then 'restore' is a misleading name. It's more like caching the data for reuse across samples, so maybe could be renamed as such.

package/com.unity.formats.usd/Runtime/Scripts/IO/SanitizedSamples.cs
```diff
@@ -273,7 +376,8 @@ internal bool ShouldUnweldVertices(bool bindMaterials)
         {
             // If any primvar is face varying (1 value per vertex) or uniform (1 value per face), all  primvars + mesh attributes will have to be converted to face varying
             // TODO: expose interpolation for standard mesh attributes (normals, tangents)
-            return normals != null && (normals.Length == originalFaceVertexCounts.Length || normals.Length > points.Length) ||
+            return arePrimvarsFaceVarying ||
```
vickycl: This is getting unwieldly and hard to parse, can we break down the logic?

vickycl (commented):
This is an awesome change but I found it hard to parse with the names chosen. I gave some suggestions that I think would make it easier to follow and therefore maintain :) 

General question- Are we generating extra Garbage with these changes?

package/com.unity.formats.usd/Runtime/Scripts/IO/SanitizedSamples.cs
mirceaispas: Because of how the assemblies are separated and used we need an extra layer of abstraction. Without this abstraction we will need to refactor how `USD.NET` and `Unity.Formats.USD` are interacting.

package/com.unity.formats.usd/Runtime/Scripts/IO/SanitizedSamples.cs
mirceaispas: The bool prevents us computing multiple times the same data.

package/com.unity.formats.usd/Dependencies/USD.NET/serialization/AccessMask.cs
vickycl: After renaming I think this should be in an IRestorableData?

package/com.unity.formats.usd/Runtime/Scripts/IO/SanitizedSamples.cs
```diff
@@ -376,17 +376,16 @@ internal bool ShouldUnweldVertices(bool bindMaterials)
...
-                tangents != null &&
-                (tangents.Length == originalFaceVertexCounts.Length || tangents.Length > points.Length) ||
-                bindMaterials &&
-                (st.GetInterpolationToken() == UsdGeomTokens.faceVarying ||
-                    uv.GetInterpolationToken() == UsdGeomTokens.faceVarying ||
-                    uv2.GetInterpolationToken() == UsdGeomTokens.faceVarying ||
-                    uv3.GetInterpolationToken() == UsdGeomTokens.faceVarying ||
-                    uv4.GetInterpolationToken() == UsdGeomTokens.faceVarying);
+            bool shouldUnweldNormals = normals != null && (normals.Length == originalFaceVertexCounts.Length || normals.Length > points.Length);
+            bool shouldUweldColors = colors != null && (colors.GetInterpolationToken() == UsdGeomTokens.uniform || colors.GetInterpolationToken() == UsdGeomTokens.faceVarying);
```
michaeljblain: Unweld

package/com.unity.formats.usd/Runtime/Scripts/IO/SanitizedSamples.cs
```diff
@@ -273,16 +376,16 @@ internal bool ShouldUnweldVertices(bool bindMaterials)
...
```
vickycl: typo: Unweld :)

package/com.unity.formats.usd/Runtime/Scripts/IO/SanitizedSamples.cs
```diff
@@ -376,17 +376,16 @@ internal bool ShouldUnweldVertices(bool bindMaterials)
...
+            bool shouldUnweldTangents = tangents != null && (tangents.Length == originalFaceVertexCounts.Length || tangents.Length > points.Length);
+            bool isTextureMappingFaceVarying = st.GetInterpolationToken() == UsdGeomTokens.faceVarying ||
+                                               uv.GetInterpolationToken() == UsdGeomTokens.faceVarying ||
+                                               uv2.GetInterpolationToken() == UsdGeomTokens.faceVarying ||
+                                               uv3.GetInterpolationToken() == UsdGeomTokens.faceVarying ||
+                                               uv4.GetInterpolationToken() == UsdGeomTokens.faceVarying;
+
+            return arePrimvarsFaceVarying || shouldUnweldNormals || shouldUweldColors || shouldUnweldTangents || (bindMaterials && isTextureMappingFaceVarying);
```
michaeljblain: ShouldUnweldColors

package/com.unity.formats.usd/Tests/USD.NET/AccessMaskTests.cs
```diff
@@ -48,22 +48,22 @@ public void ReadSample_WriteState_StateIsPreserved()
...
             scene.Read(primPath, newsample);
-            scene.AccessMask.Included[primPath].state = testState;
+            scene.AccessMask.Included[primPath].cachedData = testData;
 
             // Read a different frame and check the state
             scene.Time = 2;
             scene.IsPopulatingAccessMask = false;
             scene.Read(primPath, newsample);
-            var state = scene.AccessMask.Included[primPath].state;
+            var state = scene.AccessMask.Included[primPath].cachedData;
```
michaeljblain: For naming consistency shouldn't state be named cachedData instead

judubu:
This is great and LGTM. Thanks so much for taking it over the the finish line. 

Note that  https://github.com/Unity-Technologies/usd-unity-sdk/pull/310 into this PR (a single commit on top of the current branch) is somewhat related. This improves the loading of UVs by getting rid of the hardcoded uv attribute names and directly querying which attributes are used by the material assigned to a mesh.

vickycl (approved):
Happy to land this :)

michaeljblain (approved):

package/com.unity.formats.usd/Runtime/Scripts/IO/Geometry/MeshImporter.cs
vickycl: They are added implicitly, so this could be hit

PR #308, opened 2022-05-06
Fix regression in JointsMatch()

nicolaspopravka:
Introduced by https://github.com/Unity-Technologies/usd-unity-sdk/pull/261.

### DIFF
--- a/package/com.unity.formats.usd/Runtime/Scripts/IO/Skel/SkeletonImporter.cs
+++ b/package/com.unity.formats.usd/Runtime/Scripts/IO/Skel/SkeletonImporter.cs
@@ -159,7 +159,7 @@ namespace Unity.Formats.USD
                 return true;
             }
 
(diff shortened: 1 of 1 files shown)


vickycl:
Thank you for reporting, this has been fixed by #335 :)

nicolaspopravka:
@vickycl question for you : is the fix going to be part of an upcoming new release of the USD package ? I see that active development has resumed, which is great. So when would be a good time to upgrade ? We are still running with the "3.0.0-exp.2" release, plus the patch in the PR. Thanks

vickycl:
@nicolaspopravka hey! We're currently preparing to do a 3.0.0-exp.3 release that will include all of the latest bug fixes and improvements. We're performing some last quality passes and hope to release ASAP- can't give an exact date in case we find a major issue we need to address, but very soon :)

PR #321, opened 2022-11-28
Fixed exported Prims not having a type and fixed export for objects with same name.

mirceaispas:
## Purpose of this PR

**Ticket/Jira #:**

https://jira.unity3d.com/browse/USDU-237: Exports failing on 3.x

## Testing

Manual tests to export simple objects and objects hierarchies. A few automated tests 

I've tested with 1000 and 10000 objects the removal of Prim Cache and there is no performance degradation

## Overall Product Risks

**Complexity:**
Low

**Halo Effect:**
Low

## Additional information

**Note to reviewers:**

**Reminder:**


package/com.unity.formats.usd/Dependencies/USD.NET.Unity/UnityTypeConverter.cs
```diff
@@ -286,6 +287,33 @@ static UnityEngine.Vector3 ExtractScale(UnityEngine.Matrix4x4 mat4)
             return scale;
         }
 
+        private static bool HasAnySiblingsWithSameName(Transform transform)
+        {
```
IlliaKo, line 291: Cosmetics: you could access "name" via transform as well. Just easier to read.
Also, would be nice to just make a variable like this "var name = transform.name;" at the beginning of the function :)

IlliaKo (approved):
Verified the case with missed Mesh identifier, two other linked cases (from the same correspondent) contain the info on how to fix the other missed identifiers and no data to verify upon. So, I tried with the material missed and this situation is fixed as well. 
Will add another QA for double check

package/com.unity.formats.usd/Dependencies/USD.NET.Unity/UnityTypeConverter.cs
```diff
@@ -286,6 +287,33 @@ static UnityEngine.Vector3 ExtractScale(UnityEngine.Matrix4x4 mat4)
...
+                }
+            }
+            return false;
+        }
+
+        static public string GetUniqueName(Transform transform)
+        {
+            if (HasAnySiblingsWithSameName(transform))
+            {
+                return $"{transform.gameObject.name}{transform.GetSiblingIndex()}";
```
vickycl: Couldn't this still end up being non-unique? Say a GameObject "GO" had a sibling "GO" and a sibling "GO1", if its Sibling Index is 1 (just by chance of how they're ordered) we would get a second "GO1".

package/com.unity.formats.usd/Tests/USD.NET.Unity/UsdGeomTests.cs
```diff
@@ -179,10 +179,10 @@ public static void CameraRoundtripTest()
 
             Assert.AreEqual(unityCamOrig.orthographic, unityCamCopied.orthographic);
             Assert.AreEqual(unityCamOrig.orthographicSize, unityCamCopied.orthographicSize);
```
vickycl: Does OrthographicSize need a tolerance level too?

vickycl (commented):
Added a few comments

package/com.unity.formats.usd/Tests/Editor/ExportTests.cs
```diff
@@ -0,0 +1,102 @@
...
+            foreach (GameObject cube in cubes)
+            {
+                var cubePrim = GetPrim(cube);
+                Assert.IsNotNull(cubePrim);
+                Assert.IsTrue(cubePrim.IsValid());
+            }
+        }
+
+        [Test]
+        public void ExportMultipleSiblingGameObjectsWithSameName_AllGameObjecsHaveCorrespondingPrims()
```
vickycl: There's a small typo in these tests- "AllGameObjec**t**s"

Sorry for being really pedantic 😅

package/com.unity.formats.usd/Tests/Editor/ExportTests.cs
```diff
@@ -0,0 +1,102 @@
...
+        }
+
+        [TearDown]
+        public void TearDown()
+        {
+            m_UnityScene = EditorSceneManager.NewScene(NewSceneSetup.EmptyScene);
+            if (m_USDScene != null)
+            {
+                m_USDScene.Close();
+            }
```
vickycl, line 41: If we don't clean up USDExportTests.usda, could there be an error on a rerun of the file already existing when the setup tries to recreate it?

package/com.unity.formats.usd/Dependencies/USD.NET.Unity/UnityTypeConverter.cs
mirceaispas: Very good point, I will fix this. Thank you!

package/com.unity.formats.usd/Tests/USD.NET.Unity/UsdGeomTests.cs
mirceaispas: Yes, all floating point comparisons must be fixed. AFAIK they were fixed by https://github.com/Unity-Technologies/usd-unity-sdk/pull/323. I will rebase and if this is not the case I will fix this here

package/com.unity.formats.usd/Tests/Editor/ExportTests.cs
mirceaispas: Good point, I will check this

mirceaispas:
Rebased on dev to include CI fixes and addressed the issues found during code review

package/com.unity.formats.usd/Runtime/Scripts/IO/Scene/ImportHelpers.cs
```diff
@@ -59,7 +59,7 @@ public static GameObject ImportSceneAsGameObject(Scene scene, GameObject parent
         }
 
 #if UNITY_EDITOR
-        public static string ImportAsPrefab(Scene scene)
+        public static string ImportAsPrefab(Scene scene, string prefabPath)
```
vickycl: If this (and ImportAsTimelineClip below) are public APIs that users may already be using, I think we should give this new parameter a default of null to not give them compile errors when they upgrade

vickycl (changes requested):
Query about changing a public API

package/com.unity.formats.usd/Runtime/Scripts/IO/Scene/ImportHelpers.cs
mirceaispas: I updated the API

package/com.unity.formats.usd/Dependencies/USD.NET.Unity/UnityTypeConverter.cs
lee-aandrew: I guess in that case, the function could be named something like "HasAnySiblingsWithSameName(Transform transform)"

package/com.unity.formats.usd/Tests/Common/BaseFixture.cs
```diff
@@ -0,0 +1,120 @@
...
+
+            if (!usdFileName.EndsWith(".usd") && !usdFileName.EndsWith(".usda") && !usdFileName.EndsWith(".usdz"))
+            {
+                usdFileName += ".usda";
+            }
+
+            return Path.Combine(ArtifactsDirectoryFullPath, usdFileName);
+        }
+
+        public string GetPrefabPath(string prefabName = null)
```
lee-aandrew, line 63: Will this GetPrefabPath replace the usage of the GetPrefabPath in the ImportHelper.cs? Or are they different in usage?

vickycl (approved):
LGTM, thanks!

package/com.unity.formats.usd/Tests/Common/BaseFixture.cs
mirceaispas: This one is used only for tests - it returns a valid path in the "Artifacts" directory which will be deleted after each test.

michaeljblain (approved):
Looks good to me.

etienne-unity (approved):
Tested with various assets, comparing export between
* 2.0.0-exp.1/3.0.0-exp.1/3.0.0-exp.2/PR
In large part [USDU-237](https://jira.unity3d.com/browse/USDU-237) is fixed
_Also fixed_
[USDU-234](https://jira.unity3d.com/browse/USDU-234)  (Materials)
[USDU-235](https://jira.unity3d.com/browse/USDU-235) (Mesh)
[USDU-238](https://jira.unity3d.com/browse/USDU-238)  (export as USDZ)
_Not fixed but due to animation not importing correctly currently_
[USDU-236](https://jira.unity3d.com/browse/USDU-236) (Animation)

PR #326, opened 2022-12-09
Refactor autotests base fixture

lee-aandrew:
## Purpose of this PR

**Ticket/Jira #:**

Not in JIRA

## Testing

All UnityTestRunner returns full green for both PlayMode and EditorMode

N/A

## Overall Product Risks
Deleted USDInitilizer.cs which set the USD plugin path to "fileInfo.DirectoryName, "..", "..", "..", "Runtime", "Plugins"", but this was not the right path - but another existing USD Init script, InitUsd.cs does set the USD Plugin path to seemingly the correct path: "fileInfo.DirectoryName, "Plugins""

Not sure why USDInitializer.cs was created - would need to contact @judubu and ask

**Complexity:**
1 - refactoring

**Halo Effect:**
1

## Additional information
I worked on this on Unity version 

**Note to reviewers:**
@mirceaispas - When applicable, the tests will now inherit from BaseFixture.cs
@michaeljblain 

**Reminder:**
Talk with Julien


judubu (approved):
Good job. Thanks for cleaning that up

PR #327, opened 2022-12-15
[USDU-274] Adds new test cases for ImportHelper functions

lee-aandrew:
## Purpose of this PR
**Ticket/Jira #: https://jira.unity3d.com/browse/USDU-274**

Adds new test cases for functions within ImportHelper.cs script to increase coverage and to act as an onboarding task for @lee-aandrew 

## Testing

**Functional Testing status:**
N/A

**Performance Testing status:**
N/A

## Overall Product Risks
N/A

**Complexity:**
1

**Halo Effect:**
1

## Additional information

**Note to reviewers:**
Not sure if this is a bug or a intended way of doing things, but USD.Scene.Write, when given an object of type Parent Class, it will not write anything
For example:

SampleBase testArray[] { SphereSample sphere1, SphereSample sphere2 } // SampleBase is the root parent of SphereSample
Scene.Write("/root/sphere" testArray[0])

The Write in this case would not write anything to the scene

PR #328, opened 2022-12-22
[USDU-230][USDU-232][USDU-275] Add testcases regarding texture mapping

lee-aandrew:
## Purpose of this PR

**Ticket/Jira #:**
[USDU-230][USDU-232][USDU-275]

## Testing

Adds Testcases:
- ImportAsPrefab_TextureDataImported: Tests if Import as Prefab functionality keeps texture data set in USDA file
- ImportAsGameObjects_TextureDataImported: Tests if Import as GameObject functionality keeps texture data set in USDA file
- ImportAsGameObject_TextureDataImported_FromUsdz: Tests if Import as GameObject from USDZ file keeps texture data set in original USDA file

## Overall Product Risks
ImportHelper.cs: 
- Added another ImportAsPrefab that takes in a SceneImportOptions
- GetSelectedAssetPath: private -> public (needed access from a test script)
- GetDefaultRoot: private -> public (needed access from a test script)

**Complexity:**
1

**Halo Effect:**
1

## Additional information

**Note to reviewers:**

**Reminder:**
Requires: https://github.com/Unity-Technologies/usd-unity-sdk/pull/327 to be pushed in first


package/com.unity.formats.usd/Runtime/Scripts/IO/Scene/ImportHelpers.cs
```diff
@@ -125,7 +130,7 @@ public static string ImportAsTimelineClip(Scene scene, string prefabPath = null)
         /// <summary>
         /// Returns the selected object path or "Assets/" if no object is selected.
         /// </summary>
-        static string GetSelectedAssetPath()
+        public static string GetSelectedAssetPath()
```
vickycl: I'm not sure this API should be made public. For use in tests it could be marked `internal` instead, but making it public will encourage users to use it when it has potentially high hidden costs (Getting the whole Selection and filtering it)

package/com.unity.formats.usd/Runtime/Scripts/IO/Scene/ImportHelpers.cs
```diff
@@ -59,7 +59,7 @@ public static GameObject ImportSceneAsGameObject(Scene scene, GameObject parent
         }
 
 #if UNITY_EDITOR
-        public static string ImportAsPrefab(Scene scene, string prefabPath = null)
+        public static string ImportAsPrefab(Scene scene, SceneImportOptions importOptions, string prefabPath = null)
```
unity-chris: I don't think we should be changing public API in a test PR. afaict, it's not required for the tests, so I think we should consider doing this in a separate PR, if at all

lee-aandrew: The functionality to set the SceneImportOptions on Import is used for the testcase: ImportAsPrefab_TextureDataImported

But yes, I do agree that changing a public API in a test PR isnt very ideal
I will create a separate PR to change this part

But before that, I would also like to know, if technical design wise, this is a change that other devs can agree to (allowing custom SceneImportOptions for ImportAsPrefab function)

vickycl (changes requested):
Can we reduce the sizes or remove some of the textures please? A few 100kb can add up quite quickly over the amount of times they're downloaded, and they could be really tiny and still function the same in these tests.

package/com.unity.formats.usd/Runtime/Scripts/IO/Scene/ImportHelpers.cs
unity-chris: Aha - my search for `ImportAsPrefab(` did not find it :-) .

I think we need to do a pass of the API at some point and make these kinds of decisions, and we should add documentation etc as we go. imo this PR should add tests that validate the API/package in its current form, and we can modify/add tests later if and when we change the API.

lee-aandrew: This sounds good for me!

I will revert these changes, and for now, the testcase implementation as well - I've added a TODO tag for the case to come back to it later!

package/com.unity.formats.usd/Runtime/Scripts/IO/Scene/ImportHelpers.cs
lee-aandrew: I've reverted these changes as I no longer need it (due to: https://github.com/Unity-Technologies/usd-unity-sdk/pull/328/files#r1065024960)

lee-aandrew: I've also reduced the file sizes of the test texture files to be each less than or around 5kb

package/com.unity.formats.usd/Tests/Runtime/ImportHelpersTests.cs
```diff
@@ -58,34 +50,114 @@ public void ImportAsGameObjects_ImportAtRoot()
...
+                }
+            );
+
+            ImportAssert.IsTextureDataSaved(usdObject, fileName, isPrefab: false);
+        }
+
+        [Ignore("[USDU-275] | [USDU-230] | [FTV-202]")]
+        [TestCase(TestAssetData.FileName.TexturedOpaque, Description = "Opaque Texture")]
+        [TestCase(TestAssetData.FileName.TexturedTransparent_Cutout, Description = "Transparent Cutout Texture")]
+        public void ImportAsGameObject_TextureDataImported_FromUsdz(string fileName)
```
vickycl, line 138: I don't think USDU-275 is relevant to this test. That bug appears to be about renaming the object that contains the materials, which this test doesn't check for.

package/com.unity.formats.usd/Tests/Common/CustomAsserts/ImportAssert.cs
```diff
@@ -0,0 +1,118 @@
+using UnityEngine;
+using NUnit.Framework;
+
+namespace Unity.Formats.USD.Tests
+{
+    public static class ImportAssert
+    {
+        public static class Editor
+        {
+            enum UsdPartType
```
vickycl: I think I mentioned this on another PR so sorry if you've already addressed it, but can we avoid using 'Part' as it's not very specific and makes it harder to parse (here and below). 'Object' would probably be clearer :)

vickycl (changes requested):
Very nice! Just a few small requests: 

- Textures could still be smaller :P a single texture of a single colour shouldn't need to be more than 64x64pix, maybe even less.

- I don't see a test in here for USDU-275 (which I don't think has been fixed, so would fail anyway), am I missing something?

package/com.unity.formats.usd/Tests/Runtime/ImportHelpersTests.cs
lee-aandrew: The USDU-275 is a blocker for this test case atm due to during the assertion of this testcase, we are looking for a child named "Material" which is a child that I've hard-set in the tested .usda file

The name of the object when imported into unity is changed to "Materials" and not the expected "Material" and therefore erroring out

The error happens on:
`var materials = usdObject.transform.Find(TestAssetData.ImportGameObjectName.Material);`
in ImportAssert.IsTextureDataSaved

While the test case doesnt explicitly test this, it does require it to work as expected

But you do have a good point, I should create a separate test for USDU-275 - though this PR seems to be getting real bloated, I will do this on a separate PR

package/com.unity.formats.usd/Tests/Common/CustomAsserts/ImportAssert.cs
lee-aandrew: nono this is a issue of my messy PR requiring another PR beforehand
this has been fixed!

package/com.unity.formats.usd/Tests/Runtime/ImportHelpersTests.cs
lee-aandrew: I've also resized the textures so that they are now 16x16 pixels - they are all less than 1kb big!

vickycl (approved):
Looks good! Thank you for the iterations :)

lee-aandrew:
Thank you for the reviews @vickycl  and @unity-chris !

PR #329, opened 2023-01-04
[USDU-276] Adds testcase for Force Refreshing USDZ Imported prefab with Import Preview Surface option

lee-aandrew:
## Purpose of this PR

**Ticket/Jira #:**
[USDU-276](https://jira.unity3d.com/browse/USDU-276) 

## Testing
Adds testcase for Force Refreshing a USD prefab imported from a USDZ file, with Import Preview Surface option set for "Materials"
Is currently ignored as it will cause a failure due to [USDU-276](https://jira.unity3d.com/browse/USDU-276) 

## Overall Product Risks

**Complexity:**
1

**Halo Effect:**
1


vickycl (changes requested):
I'm not entirely clear what the goal of this test is. I think this test (and the bug) are more about the ImportPreviewSurface than the reload behaviour, is that correct? If so, I think this test should be moved to a material specific file, and/or make that more clear in the test. In its current state it's not clear to me what additional value this test adds.

lee-aandrew:
> I'm not entirely clear what the goal of this test is. I think this test (and the bug) are more about the ImportPreviewSurface than the reload behaviour, is that correct? If so, I think this test should be moved to a material specific file, and/or make that more clear in the test. In its current state it's not clear to me what additional value this test adds.

Hey Vicky, the goal of this test is to make sure that error handling is done correctly when Refresh errors out with the testing circumstances stated during the test - and as far as i know the steps taken in the test case is the only way a user may refresh a usdz imported prefab with Import Preview Surface without having to bring it into the Unity Scene

The same error does not happen when you directly import it as a prefab with the Scene Import Options with materials set to ImportPreviewSurface (have to temporarily change setting in ImportHelpers.ImportAsPrefab) so it seems like a problem with the Refresh feature more than specifically the ImportPreviewSurface setting

Though I do agree I havent been too descriptive in the naming of the test case, I changed it to 
"UsdAsset_Force_Reload_Prefab_ImportPreviewSurface_ErrorShouldCleanScene"
But let me know if you still dont get the value of the test case in first glance - in which case i will need to do something more to it

package/com.unity.formats.usd/Tests/Editor/UsdAssetReloadTests.cs
```diff
@@ -109,6 +121,26 @@ public void UsdAsset_Reload_FileDidNotChange_ValuesDoNotChange()
...
+        {
+            var initialRootGameObjects = UnityEngine.SceneManagement.SceneManager.GetActiveScene().GetRootGameObjects();
+            var scene = OpenUSDGUIDAssetScene(k_USDZGUID, out _);
+            prefabUsdPath = ImportHelpers.ImportAsPrefab(scene);
+            scene.Close();
+
+            var prefabAsset = AssetDatabase.LoadAssetAtPath<UsdAsset>(prefabUsdPath);
+            prefabAsset.m_materialImportMode = MaterialImportMode.ImportPreviewSurface;
+
+            LogAssert.Expect(LogType.Error, "Destroying assets is not permitted to avoid data loss.\nIf you really want to remove an asset use DestroyImmediate (theObject, true);");
```
vickycl: I don't think this error should be expected, so I think the test should fail if it's thrown

vickycl (changes requested):
DM'd on Slack, I think this test should probably be two based on reproducing the original bug :)

package/com.unity.formats.usd/Tests/Editor/UsdAssetReloadTests.cs
lee-aandrew: Changed it so that it no longer expects an error on force reload
also separated the tests into two tests:
- Expects Success without errors for all 3 MaterialImportModes
- Expects Scene is not dirtied on Import Preview Surface mode force reload

vickycl (approved):
These new tests look much clearer :)

lee-aandrew:
Thanks for the reviews @vickycl !

PR #330, opened 2023-01-04
[USDU-249] Adds test case for InitUSD

lee-aandrew:
## Purpose of this PR

[USDU-249](https://jira.unity3d.com/browse/USDU-249)

## Testing
Adds test cases for InitUsd script
tries to check if an early throw / exit is done when given wrong parameter value for function InitUsd.SetupUsdPath()

**Complexity:**
1

**Halo Effect:**
1

### DIFF
--- a/package/com.unity.formats.usd/Tests/Editor/InitUsdTests.cs
+++ b/package/com.unity.formats.usd/Tests/Editor/InitUsdTests.cs
@@ -1,0 +1,73 @@
+using System.Collections;
+using System.IO;
+using NUnit.Framework;
+using UnityEditor;
+using UnityEngine;
+using UnityEngine.TestTools;
+using USD.NET;
+using USD.NET.Unity;
+using System.Reflection;
+using Unity.Formats.USD;
+using System.Runtime.CompilerServices;
+
+public class InitUsdTests
+{
+    [Test]
+    [Ignore("[USDU-249]")]
+    public void SetupUsdPath_InvalidPath_Error()
+    {
+        // SetupUsdPath function is a private function
+        var setUpMethod = GetMethod("SetupUsdPath");
+
+        try
+        {
+            setUpMethod.Invoke(null, new object[] { "\\NonExisting\\Path\\" });
+        }
+        catch (TargetInvocationException e)
+        {
+            Assert.AreEqual(e.InnerException.GetType(), typeof(FileNotFoundException));
+            return;
+        }
+
+        Assert.Fail("Exception was expected but none was thrown");
+    }
+
+    [Test]
+    public void SetupUsdPath_EmptyPath_Error()
+    {
+        // SetupUsdPath function is a private function
+        var setUpMethod = GetMethod("SetupUsdPath");
+
+        try
+        {
+            setUpMethod.Invoke(null, new object[] { "" });
+        }
+        catch (TargetInvocationException e)
+        {
+            Assert.AreEqual(e.InnerException.GetType(), typeof(System.ArgumentException));
+            return;
+        }
+
+        Assert.Fail("Exception was expected but none was thrown");
+    }
+
+    [Test]
+    public void InitUsd_Initialize()
+    {
+        // Reset 'm_usdInitialized' for accurate testing
+        var isUsdInitialized = typeof(InitUsd).GetField("m_usdInitialized", (BindingFlags.Static | BindingFlags.NonPublic));
+        isUsdInitialized.SetValue(null, false);
+
+        Assert.True(InitUsd.Initialize());
+    }
+
+    private MethodInfo GetMethod(string methodName)
+    {
+        var method = typeof(InitUsd).GetMethod(methodName, (BindingFlags.NonPublic | BindingFlags.Static));
+
+        if (method == null)
+            Assert.Fail(string.Format("{0} method not found", methodName));
+
+        return method;
+    }
+}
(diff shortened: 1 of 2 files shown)


package/com.unity.formats.usd/Tests/Editor/InitUsdTests.cs
```diff
@@ -0,0 +1,74 @@
...
+        // SetupUsdPath function is a private function
+        var setUpMethod = GetMethod("SetupUsdPath");
+
+        try
+        {
+            setUpMethod.Invoke(null, new object[] { "\\NonExisting\\Path\\" });
+        }
+        catch (TargetInvocationException e)
+        {
+            Assert.AreEqual(e.InnerException.GetType(), typeof(FileNotFoundException));
```
unity-chris: Should we use Assert.Throws for these and check the messages?

lee-aandrew: I would like to do it this way as well, but im not an expert when it comes to Reflection, and im not sure how to get the data of the thrown error

Invoke returns the value of the invoked method, in this case, void

To detect that an error was thrown by the invoked method, and then get additional info about the thrown exception, i can't really think of a way to do it without putting it under a try, catch statement

unity-chris: You should be able to do something like this:

```
var ex = Assert.Throws<TargetInvocationException>(() => setUpMethod.Invoke(null, new object[] {""}));
Assert.IsInstanceOf<System.ArgumentException>(ex.InnerException);
Assert.That(ex.InnerException.Message, Is.EqualTo("The specified path is not of a legal form (empty)."));
```

lee-aandrew: Cool i didnt know assert.throws returns the error data!
Thank you Chris!

package/com.unity.formats.usd/Tests/Editor/InitUsdTests.cs
```diff
@@ -0,0 +1,61 @@
...
+        // SetupUsdPath function is a private function
+        var setUpMethod = GetMethod("SetupUsdPath");
+
+        var expectedError = Assert.Throws<TargetInvocationException>(() => setUpMethod.Invoke(null, new object[] { "" }), "Error was expected but not thrown");
+        Assert.IsInstanceOf<System.ArgumentException>(expectedError.InnerException);
+        Assert.AreEqual("The specified path is not of a legal form (empty).", expectedError.InnerException.Message, "Unexpected error message was given");
+    }
+
+    [Test]
+    public void InitUsd_Initialize()
```
vickycl, line 35: Please add the expected result to the end of the test name, something like "InitUsd_Initialize_Succeeds"

package/com.unity.formats.usd/Tests/Editor/InitUsdTests.cs
```diff
@@ -0,0 +1,61 @@
...
+    {
+        // Reset 'm_usdInitialized' for accurate testing
+        ResetInitUsd();
+
+        Assert.True(InitUsd.Initialize(), "USD Initialize failed");
```
vickycl, line 40: Is there a more robust thing we can also check for, as well as the return value of Initialise()?

vickycl: eg check that the plugin is registered, that the expected debug log is printed, that no other errors are handled and logged.

package/com.unity.formats.usd/Tests/Editor/InitUsdTests.cs
```diff
@@ -0,0 +1,61 @@
...
+        var expectedError = Assert.Throws<TargetInvocationException>(() => setUpMethod.Invoke(null, new object[] { "" }), "Error was expected but not thrown");
+        Assert.IsInstanceOf<System.ArgumentException>(expectedError.InnerException);
+        Assert.AreEqual("The specified path is not of a legal form (empty).", expectedError.InnerException.Message, "Unexpected error message was given");
+    }
+
+    [Test]
+    public void InitUsd_Initialize()
+    {
+        // Reset 'm_usdInitialized' for accurate testing
+        ResetInitUsd();
```
vickycl, line 38: I'm not sure that just changing the bool is enough here. Initialize() does stuff that might affect this test, such as setting up the UnityTypeBindings for USD. I'm not sure what the answer is though :D

vickycl (changes requested):
I think we need to rethink how to thoroughly test the InitUSD.Initialize test without it impacting/ being impacted by the other tests having the plugin initialised. Perhaps we could make a fake instance (not sure how it would interact with the real instance though) or maybe we need a test outside of UTR...

package/com.unity.formats.usd/Tests/Editor/InitUsdTests.cs
lee-aandrew: i've looked into this a bit more - on how to tell if the plugin has been registered and the typebinding
But I just ended up having more questions, do you know who I should be asking about the init steps?

judubu: You can check that unity types have been added the TypeBinder, that usd error message are logged in Unity console and that the plugins Registry is not empty. We can chat if you have more questions.

lee-aandrew: Thank you @judubu - I will contact you via slack!

lee-aandrew: After talking with Julien, it turns out the PlugRegistry is not behaving correctly - created a separate JIRA ticket for this
https://jira.unity3d.com/browse/USDU-295

Will continue work on this PR after the above JIRA ticket has been solved

PR #331, opened 2023-01-10
[USDU-199] Delete GameObjects when updating USD Instancers to avoid duplication.

### LINKED ISSUE #72: Instanceable true on reference breaks mesh (FTV-206)
When m_usdTimeOffset is set during runtime, referenced meshes with `instanceable=true` set, break and are randomly duplicated throughout the scene and usd hierarchy.

Examples can be provided. 

Have successfully removed `instanceable=true` and have fixed the breaking meshes.

Breaks mesh on m_usdTimeOffsetChange
``` 
   def Xform "domeHero_001" (
        instanceable = true
        prepend references = @/path/to/asset.usd@
    )
    {
        float xformOp:rotateY = -267.4804
        double3 xformOp:translate = (0, 1128.6803799705747, 0)
        uniform token[] xformOpOrder = ["xformOp:translate", "xformOp:rotateY"]
    }
```
Acts as intended

``` 
   def Xform "domeHero_001" (
        prepend references = @/path/to/asset.usd@
    )
    {
        float xformOp:rotateY = -267.4804
        double3 xformOp:translate = (0, 1128.6803799705747, 0)
        uniform token[] xformOpOrder = ["xformOp:translate", "xformOp:rotateY"]
    }

### PR DESCRIPTION
**Pull request by vickycl**
## Purpose of this PR

**Ticket/Jira #:** [USDU-199](https://jira.unity3d.com/browse/USDU-199?filter=-3), [USDU-204](https://jira.unity3d.com/browse/USDU-204?filter=-1)

When refreshing a USD asset, Point Instancers were generating a new set of instances without cleaning up the previous ones, resulting in duplication. This change destroys the old instances first to avoid the duplication.

Performance wise this isn't great, as destroying GameObjects has a cost. However (I don't think) we have a way of knowing whether the prototype has changed, so this seems like the safest way for now. This hit will only be on importing a large number of point instances.

I've also added a fix for the error `ApplicationException: USD ERROR: /instancer -- all sample times in times and baseTime must either all be numeric or all be default` that was occurring on refresh of any PointInstancer. I've based the fix on the [docs](https://graphics.pixar.com/usd/dev/api/class_usd_geom_point_instancer.html#a619b5f21f6c0fc26e6bd25c86c82b390):

> If your application does not care about off-sample interpolation, it can supply the same value for baseTime that it does for time.

I'm assuming for the time being that we do not care about this.

## Testing

Tested against original asset (linked in Jira), confirmed no more duplication.

Measured the performance of previous vs new with Superluminal- for 100 instances, before = 17ms, after = 22ms. Average of 0.05ms extra per instance. 

## Overall Product Risks

**Complexity:** Low

**Halo Effect:** Low

## Additional information

**Note to reviewers:**

**Reminder:**

### CODE REVIEW

File: package/com.unity.formats.usd/Runtime/Scripts/IO/Geometry/InstanceImporter.cs [hunk 1]
```diff
@@ -126,11 +129,23 @@ public static void BuildSceneInstances(PrimMap primMap, SceneImportOptions optio
                 }
 
                 var xf = transforms[i];
+
+                var instanceName = $"{goMaster.name}_{i}";
+
+                // If the old instance exists, we must destroy it to avoid a duplicate
```
lee-aandrew, line 135: since this commented section is repeated twice, could we turn this into a small function instead?

vickycl: I think they're different enough to be left alone (one looks at the existing instance, and one at its child) and the resulting function would only be like 2 lines long. Personally I stick to rule of only extracting to a separate function if there are 3 repeats, but I guess that's a personal preference :)

lee-aandrew:
LGTM!

lee-aandrew (approved):

mirceaispas (approved):

package/com.unity.formats.usd/CHANGELOG.md
```diff
@@ -4,6 +4,7 @@
 ### Bug Fixes
 - Fixed an import bug causing instanced primitives not to be sanitized.
 - Disabled plugins on unsupported platforms.
+- Fixed a bug causing PointInstances to be duplicated.
```
michaeljblain, line 7: Might be worth also adding the time fix to the changelog

michaeljblain (changes requested):

package/com.unity.formats.usd/CHANGELOG.md
vickycl: Good point, thanks!

michaeljblain: Can we include some of the text of the exception itself so a user can find that in the CHANGELOG?

vickycl: Done :)

michaeljblain (approved):

PR #332, opened 2023-01-10
[USDU-199] Add Test case for Point Instancing Fix

lee-aandrew:
## Purpose of this PR

**Ticket/Jira #:**

[USDU-199](https://jira.unity3d.com/browse/USDU-199?filter=-3)

## Testing

UsdInstancerAsset_Reload_NoDuplicateObjects:
Imports a USD instancer object as a GameObject into Unity, checks if USD Refresh (force = false) deletes all existing instancers before creating new ones
Checks for object count and that the new object does not have the same InstanceID as the old one to ensure deletion has been done

Adds about 0.3 seconds to our current test suite

## Overall Product Risks

**Complexity:**
1

**Halo Effect:**
1

## Additional information

**Note to reviewers:**

**Reminder:**


package/com.unity.formats.usd/Tests/Editor/UsdAssetReloadTests.cs
```diff
@@ -109,6 +117,38 @@ public void UsdAsset_Reload_FileDidNotChange_ValuesDoNotChange()
...
+        [Test]
+        [Ignore("USDU-199")]
+        public void UsdInstancerAsset_Reload_NoDuplicateObjects()
+        {
+            var instancerChildNamePrefix = "Root";
+
+            var scene = OpenUSDGUIDAssetScene(k_USDInstancerGUID, out _);
+            var instancerObject = ImportHelpers.ImportSceneAsGameObject(scene);
+
+            var originalChildrenCount = instancerObject.transform.childCount;
```
vickycl: I think it would be good to store the expected value from the USDA as a const in the test (I believe the file defines 100 point instances), else if something broke and the value was wrong (zero, for example) this test would miss it.

vickycl (changes requested):
This PR looks like it contains content from your previous test PRs, I guess because they need some of the same setup code. It does make it quite hard to read, and could cause merge conflicts and things to get broken if changes are made on some of the branches that land before this one. It should be fine this time, but take care when merging these :) 

Overall I think the instancing test is good, just needs one small change to store the data and save it from potentially missing a breakage in future.

package/com.unity.formats.usd/Tests/Editor/UsdAssetReloadTests.cs
lee-aandrew: sorry about the messy PR, i've rebased it with the new updated origin/dev so it should be cleaner

also, thats a good point on the expected child count - I've set it as a const int of 101 // 100 instancers + "Prototypes" object

vickycl (approved):
Great work! :)

PR #334, opened 2023-01-13
[Draft] Fix for JointsMatch

vickycl:
## Purpose of this PR

**Ticket/Jira #:** PR #308 

A user pointed out that JointsMatch() functionality regressed in PR #261, resulting in it returning true in nearly all cases. 

I'm not sure whether the intended functionality for this method is to just check whether the two inputs are the same instance, or whether the full array of strings is the same, but as the behavior prior to the change was the latter I have reverted it to do that.

## Testing

Not sure how best to test this as no repro case was supplied. I'll run the normal test suite.

Very small performance impact expected due to the amount of string comparisons that may be introduced.

## Overall Product Risks

**Complexity:** minimal

**Halo Effect:** minimal

## Additional information

**Note to reviewers:**

**Reminder:**

### DIFF
--- a/package/com.unity.formats.usd/Runtime/Scripts/IO/Skel/SkeletonImporter.cs
+++ b/package/com.unity.formats.usd/Runtime/Scripts/IO/Skel/SkeletonImporter.cs
@@ -159,6 +159,15 @@ namespace Unity.Formats.USD
                 return true;
             }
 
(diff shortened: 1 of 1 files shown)


vickycl:
(turns out this is a duplicate effort)

PR #338, opened 2023-01-18
[USDU-260] Fix 'ExportSelectedWithChildren' on Mac

vickycl:
## Purpose of this PR

**Ticket/Jira #:** [USDU-260](https://jira.unity3d.com/browse/USDU-260)

MacOS doesn't support passing multiple file extensions to EditorUtility.SaveFilePanel and so was appending all of them. This changes the 'Export Selected With Children' button on non-Windows platform to default to .usd, and adds a 'Export Selected With Children as USDA' option for the alternative.

Screenshot 2023-01-27 at 15 19 09

## Testing

Manually tested that the bug no longer reproduces, and looked at the resulting .usd in USDView to confirm the output was as expected.

No anticipated performance impact.

## Overall Product Risks

**Complexity:** low

**Halo Effect:** lox

## Additional information

**Note to reviewers:**

**Reminder:**


SimonBoorer:
You could also add two options on non-Windows platforms (e.g. Export USD ASCII File, Export USD Crate File)

SimonBoorer (approved):

michaeljblain (approved):

package/com.unity.formats.usd/Editor/Scripts/Behaviors/UsdMenu.cs
```diff
@@ -23,6 +23,7 @@ namespace Unity.Formats.USD
     public static class UsdMenu
     {
         [MenuItem("USD/Export Selected with Children", true)]
+        [MenuItem("USD/Export Selected with Children as USDA", true)]
```
SimonBoorer, line 26: Is it OK that this is defined on Windows but the actual `MenuExportSelectedWithChildrenToUSDA` is not?

SimonBoorer (approved):

package/com.unity.formats.usd/Editor/Scripts/Behaviors/UsdMenu.cs
vickycl: From first read of the docs it sounded like it was fine but looking at the code it seems like it could be a weird edge case. Although it works as expected on Windows currently I'm not confident it won't break in future. I will change it :)

PR #339, opened 2023-01-19
[USDU-275] Adds testcase for checking if export keeps the original data structure

lee-aandrew:
## Purpose of this PR

**Ticket/Jira #:**
[[USDU-275](https://jira.unity3d.com/browse/USDU-275)] & [[USDU-279](https://jira.unity3d.com/browse/USDU-279)]

## Testing
ExportAsUsdz_DataStructureKeptOnImport:
1. Imports a known USDA file
2. Exports it as a UDSZ file
3. Import the USDZ file
4. Check if the imported USDZ file maintains the same game object structure as the initially imported USDA file

**Complexity:**
1

**Halo Effect:**
1

## Additional information

**Note to reviewers:**

**Reminder:**

### DIFF
--- a/package/com.unity.formats.usd/Tests/Common/CustomAsserts/ExportAssert.cs
+++ b/package/com.unity.formats.usd/Tests/Common/CustomAsserts/ExportAssert.cs
@@ -1,0 +1,34 @@
+using UnityEngine;
+using NUnit.Framework;
+
+namespace Unity.Formats.USD.Tests
+{
+    public static class ExportAssert
+    {
+        public static class Runtime
+        {
+            public static void IsDataStructureKeptInUsdz(GameObject importedObject, string expectedObjectName, string expectedRootPrimName, string expectedMaterialsName)
+            {
+                Assert.AreEqual(importedObject.name, expectedObjectName);
+
+                var rootPrim = importedObject.transform.Find(expectedRootPrimName);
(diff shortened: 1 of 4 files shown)


package/com.unity.formats.usd/Tests/Common/CustomAsserts/ExportAssert.cs
```diff
@@ -0,0 +1,34 @@
+using UnityEngine;
+using NUnit.Framework;
+
+namespace Unity.Formats.USD.Tests
+{
+    public static class ExportAssert
+    {
+        public static class Runtime
```
vickycl: I don't think this extra layer of breakdown is necessary, also because nothing in this function is specific to runtime?

vickycl (commented):
Is there a reason for making this test a Runtime test? I can't think of a reason not to, but it does seem like exporting in EditMode is a far more common use case, so I'm curious :)

package/com.unity.formats.usd/Tests/Common/CustomAsserts/ExportAssert.cs
lee-aandrew: You have a good point, I originally put it here to be with the other ExportHelper tests, but I moved it to the EditMode ExportTests, but expanding on that, I think we can move a lot of export and import tests to EditorMode tests as well

PR #341, opened 2023-01-26
[USDU-219] Mark UVs attributes as varying interpolation

### LINKED ISSUE #284: USD export using incorrect texture mapping  interpolation option on primvars:st attribute (USDU-219)
When exporting an object with texture mapping and thus st coordinates an incorrect option is chosen for the `primvars:st` attribute. This causes the texture mapping to not happen. Manually changing the "constant" to "varying" corrects the problem.

This does not show a texture:
```
        def Mesh "H_DDS_LowRes"
        {
            uniform bool doubleSided
             ...
           float2[] primvars:st = [(0.90359735, 0.9734193), .... ((0.2584137, 0.4466889)] (
                elementSize = 1
                interpolation = "constant"
            )
           ...
```

This however does :

```
        def Mesh "H_DDS_LowRes"
        {
            uniform bool doubleSided
             ...
           float2[] primvars:st = [(0.90359735, 0.9734193), .... ((0.2584137, 0.4466889)] (
                elementSize = 1
                interpolation = "varying"
            )
           ...
```

Notes:
- Using Unity 2020.3.20f1
- Using USD Package 3..0.0-exp.2
- A repo with Python code that does a minimal parsing of a USDA file and repairs the issues 278, 282, 283, 284, 285  can be found at `https://github.com/MikeWise2718/usdmorph`  - hopefully it won't be needed long before these are all repaired in a new version.

vickycl:
## Purpose of this PR

**Ticket/Jira #:** [USDU-219](https://jira.unity3d.com/browse/USDU-219?filter=-1), #284 

Sets PrimVar interpolation for texture mapping on exported meshes to varying. In Unity every attribute is per vertex, so this default makes sense.

## Testing

Confirmed manually that the issue (texture mapping not happening) no longer reproduces by exporting a mesh with a mapped texture and looking at the result in UsdView. Adds a new test that exports a mesh, then imports it again to check the PrimVar set.

No anticipated performance impact.

## Overall Product Risks

**Complexity:** low

**Halo Effect:** low

## Additional information

**Note to reviewers:**

**Reminder:**


package/com.unity.formats.usd/Tests/Editor/ExportTests.cs
```diff
@@ -117,5 +117,23 @@ public void ExportMultipleSiblingGameObjectsWithSameName_AllGameObjectsHaveCorre
...
+
+            ExportHelpers.ExportGameObjects(new GameObject[] { cube }, ExportHelpers.InitForSave(m_USDScenePath), BasisTransformation.SlowAndSafe);
+
+            m_USDScene = Scene.Open(m_USDScenePath);
+            pxr.UsdPrim cubePrim = GetPrim(cube);
+
+            pxr.UsdGeomMesh usdGeomMesh = new pxr.UsdGeomMesh(cubePrim);
+            var stPrimvar = usdGeomMesh.GetPrimvar(new pxr.TfToken("st"));
+            Assert.IsNotNull(stPrimvar, $"Mesh {cube.name} has no 'st' primvar.");
+            Assert.AreEqual(new pxr.TfToken("varying"), stPrimvar.GetInterpolation(), $"st on mesh {cube.name} is not set to varying interpolation.");
```
judubu: We have the common geometry tokens defined as static TfTokens in UsdGeomTokens. You could use UsdGeomTokens.varying here

judubu (approved):
One minor comment in the test, but I'd say ship it.

package/com.unity.formats.usd/Tests/Editor/ExportTests.cs
vickycl: Oh cool, I didn't know about those, thanks :) Is it worth generating one for st in there too?

judubu: UsdGeomTokens contains the token values defined by the  various UsdGeom schemas (see https://github.com/Unity-Technologies/usd-unity-sdk/blob/dev/package/com.unity.formats.usd/Runtime/Plugins/x86_64/usd/usdGeom/resources/generatedSchema.usda). And we also have UsdShadeTokens.cs and UsdSkelTokens.cs to store other schema defined tokens. 

If we want static Tokens for our own needs we should add them in a different file.

SimonBoorer (approved):

michaeljblain (approved):
Looks good to me.

jcowles:
This is awesome! 🎉 🎉 🎉

PR #342, opened 2023-01-30
[USDU-245] Adds Test case for EditorOnly tag object not being exported

lee-aandrew:
## Purpose of this PR

**Ticket/Jira #:**

[USDU-245](https://jira.unity3d.com/browse/USDU-245)
Objects tagged as "EditorOnly" are still exported #303

## Testing

ExportObjectWithEditorOnlyTag_DoesNotExportEditorOnly
1. Create a root object
2. Creates a default object - set it as 1st child of root object
3. Creates an EditorOnly Tag object - set it as 2nd child of root object
4. Exports it as a USD file
5. Open and iterate over the exported USD file
6. Make sure the EditorOnly Tag child object does not exist

**Complexity:**
1

**Halo Effect:**
1

## Additional information

**Note to reviewers:**

Currently set as Ignore due to [USDU-245](https://jira.unity3d.com/browse/USDU-245)
Will remove the ignore attribute once ticket is solved

**Reminder:**


package/com.unity.formats.usd/Tests/Editor/ExportTests.cs
```diff
@@ -117,5 +117,34 @@ public void ExportMultipleSiblingGameObjectsWithSameName_AllGameObjectsHaveCorre
...
+            m_USDScene = Scene.Open(m_USDScenePath);
+
+            foreach (Transform child in rootObject.transform)
+            {
+                if (child.tag == editorOnly)
+                {
+                    var childPrim = GetPrim(child.gameObject);
+
+                    Assert.IsNull(childPrim, $"GameObject <{child.name}> with Tag 'EditorOnly' Shouldn't have been exported");
+                }
```
vickycl: I think this should also test that there is not a game object present with the name "EditorOnlyTag_Object", incase the object is exported anyway just without the tag.

vickycl (changes requested):
Small addition requested

package/com.unity.formats.usd/Tests/Editor/ExportTests.cs
lee-aandrew: good point, will add!

lee-aandrew: I changed the test case so it doesnt specifically do the assertion once an object with "EditorOnly" tag has been found - it will just look for the actual object that shouldnt be there

vickycl (approved):
LGTM

lee-aandrew:
Thanks for the reviews @vickycl !

PR #343, opened 2023-01-30
[USDU-280] Reduced the chance of import timing out for complex meshes.

vickycl:
## Purpose of this PR

**Ticket/Jira #:** [USDU-280](https://jira.unity3d.com/browse/USDU-280) / #324

For certain mesh setups (eg where st and normals attributes are of faceVarying interpolation) we have to unweld every face of the mesh to get consistent array sizes. During this sanitization we can timeout and report an error if it takes >1s.

- Increased the timeout on waiting for the read job from 1 second to 10 seconds
- Optimized the Triangulate() function to remove dynamically sized Lists, reducing allocations.

## Testing

Tested that the original timeout no longer reproduces.

This PR halved the time taken for triangulation in my local testing, profiled with Superluminal:

Before:
image

After:
image

## Overall Product Risks

**Complexity:** Low

**Halo Effect:** Low

## Additional information

**Note to reviewers:**

@judubu please advise of any further situations I should test?

**Reminder:**


package/com.unity.formats.usd/Runtime/Scripts/IO/ReadJob.cs
```diff
@@ -140,7 +140,7 @@ public bool MoveNext()
                 }
 
                 j++;
-                if (!m_ready.WaitOne(1000))
+                if (!m_ready.WaitOne(10000))
```
lee-aandrew, line 143: Just a question, is the increase in time out still intentional with the optimization?
If so, is there a need for a discussion on the right value for a time out, and if there is also a need for something like a "cancel import"  button during the import load time?

package/com.unity.formats.usd/CHANGELOG.md
```diff
@@ -6,6 +6,7 @@
 - Disabled plugins on unsupported platforms.
 - Fixed a bug causing PointInstances to be duplicated.
 - Fixed a bug causing an ApplicationException "all sample times in times and baseTime must either all be numeric or all be default" to be thrown when refreshing a PointInstancer.
+- Optimized triangulation to reduce the chance of a "Timed out while waiting for thread" error on importing a complex mesh.
```
lee-aandrew, line 9: I did a comparison between the imports pre and post this change, and with my sample file, there was a reduction from 10 seconds to 5 seconds, which seems like a pretty big cut down in time and matches with Vicky's local testing

My test file has:
Batches: 42
Tris: 232.8M
Verts: 116.4M

judubu (approved):
Awesome gains. Thanks for optimizing.

package/com.unity.formats.usd/Runtime/Scripts/IO/ReadJob.cs
vickycl: Yeah, it's still needed. I gambled with 10s being an 'acceptable' timeout given that the Triangulate code is twice as fast. Theoretically a file 20x the size of the repro case will hit this same thing, but given the specific conditions that have triggered it in this case (st and normals set to faceVarying) and the easy workaround for users encountering this in future (increase the timeout) I think this is acceptable _for now_. In future we should definitely look at a more thorough fix, and a cancel option would be very valid. @SimonBoorer mentioned the possibility of using Hydra's triangulate capability, I don't know if that would necessarily be faster but that's a little way off yet.

lee-aandrew: Thank you for the explanation!

lee-aandrew (approved):

package/com.unity.formats.usd/Runtime/Scripts/IO/Geometry/MeshImporter.cs
```diff
@@ -466,7 +466,12 @@ public static void ImportSkinning(string path, SanitizedMeshSample usdMesh, Mesh
                     var newFaceIndices = new List<int>();
                     foreach (var faceIndex in nameAndSubset.Value)
                     {
-                        newFaceIndices.AddRange(usdMesh.faceMapping[faceIndex]);
+                        int firstFace = usdMesh.faceMapping[faceIndex];
+                        int faceCount = usdMesh.originalFaceVertexCounts[faceIndex] - 2;
```
michaeljblain: Super minor: change variable name to faceVertexCount to align with originalFaceVertexCounts

package/com.unity.formats.usd/Runtime/Scripts/IO/SanitizedSamples.cs
```diff
@@ -200,52 +201,71 @@ internal void Triangulate(bool changeHandedness)
...
             int currentFaceOffset = 0, currentIndexOffset = 0;
-            for (var i = 0; i < faceVertexCounts.Length; i++)
-            {
-                // This could be faster if we had a fixed vertex count so we could allocate one single large array
-                faceMapping[i] = new int[faceVertexCounts[i] - 2];
 
-                var next = last + 1;
-                var t = 0;
-                for (; t < faceVertexCounts[i] - 2; t++)
+            if (changeHandedness)
```
michaeljblain, line 224: Would be good to state which handedness we're changing from/to

michaeljblain (changes requested):
Overall looks good. Some minor suggestions.

package/com.unity.formats.usd/Runtime/Scripts/IO/Geometry/MeshImporter.cs
vickycl: It isn't actually the vertex count, it is the number of faces generated by the triangulation. So for example a quad face (vertexCount = 4) generates 2 triangle faces (vertexCount - 2). I'll change it to  triCount and add a comment :)

michaeljblain: Ah yes triCount is perfect.

package/com.unity.formats.usd/Runtime/Scripts/IO/SanitizedSamples.cs
vickycl: I added a comment- does it make sense?

michaeljblain: Looks good.

michaeljblain (approved):

PR #344, opened 2023-01-30
[USDU-292] Adds testcase to check physical camera data on export and import

lee-aandrew:
## Purpose of this PR

**Ticket/Jira #:**

[USDU-292](https://jira.unity3d.com/browse/USDU-292)

## Testing
ExportPhysicalCamera_RetainsPhysicalRelatedData
1. Adds Camera component to a gameObject
2. Set Physical Camera setting to True
3. Set focalLength, sensorSize and lensShift
4. Export the camera object
5. Ensure the exported camera object has physical camera data within its USD properties

ImportPhysicalCamera_PhysicalDataKept
1. Created a Physical camera object on Blender and exported as USDA file (physicalCam.usda)
2. Do Assertion on physical camera properties set in usda file

**Complexity:**
1

**Halo Effect:**
1

## Additional information

**Note to reviewers:**

**Reminder:**


package/com.unity.formats.usd/Tests/Editor/ImportTests.cs
```diff
@@ -0,0 +1,40 @@
+using System.IO;
+using NUnit.Framework;
+using pxr;
+using UnityEditor;
+using UnityEngine;
+using USD.NET;
+
+namespace Unity.Formats.USD.Tests
+{
+    public class ImportTests : BaseFixtureEditor
```
vickycl: Could this just be `public class ImportCameraTests : BaseFixtureEditor`? I don't think there's any benefit to this additional inheritance

package/com.unity.formats.usd/Tests/Editor/ImportTests.cs
```diff
@@ -0,0 +1,40 @@
...
+    {
+        public class CameraRelated : ImportTests
```
vickycl: I think 'CameraRelated' is vague, maybe something like 'ImportCameraTests'? This would be in line with the other tests in this folder eg ImportMeshTests

package/com.unity.formats.usd/Tests/Editor/ExportTests.cs
```diff
@@ -117,5 +116,46 @@ public void ExportMultipleSiblingGameObjectsWithSameName_AllGameObjectsHaveCorre
             }
             Assert.AreEqual(cubes.Length, exportedPrims.Count, "One or more GameObjects don't have a corresponding Prim");
         }
+
+        class CameraRelated : USDExportTests
```
vickycl: Out of curiosity, what's the benefit of this nested class/ inheritance structure over just having this lass outside of USDExportTests? I think  'ExportCameraTests' or some such might be a clearer name too :)

vickycl (changes requested):
Nice tests! I think 'CameraRelated' is a bit vague and should just be 'Camera' in the folder structures and 'ExportCameraTests'/ 'ImportCameraTests' inline with the other tests. Also just some queries about inheritance/ nesting, but overall looks good :)

package/com.unity.formats.usd/Tests/Editor/ExportTests.cs
lee-aandrew: I wanted to sub-categorize / sub-group the test into a more specific part within the USDExportTests, but yeah, I guess I could just have them as a separate class all together instead - @IlliaKo also mentioned something similar at a call today as well

package/com.unity.formats.usd/Tests/Common/TestUtilityFunction.cs
```diff
@@ -0,0 +1,34 @@
...
+        {
+            var usdPath = Path.GetFullPath(AssetDatabase.GUIDToAssetPath(usdGUID));
+            var stage = UsdStage.Open(usdPath, loadSet);
+            var scene = Scene.Open(stage);
+
+            return scene;
+        }
+    }
+#endif
+    public static class ExportTest
```
vickycl: This isn't export specific, so I think this extra layer of abstraction will make it hard to find this for reuse. I think both of these functions can just be inside 'TestUtilityFunctions'

vickycl (changes requested):
One small change requested but otherwise looking better :)

vickycl (approved):
Great, thanks for the iterations!

lee-aandrew:
Thank you for the reviews! @vickycl !

PR #346, opened 2023-02-02
Optimized PointInstance importing

vickycl:
## Purpose of this PR

**Ticket/Jira #:** N/A

PR #331 introduced lots of Transform.Find calls when importing a PointInstancer, which in initial testing didn't appear to introduce a regression but has since proved to be problematic. Discussing with @John-Whigham, we discovered this is because the order of the transform hierarchy means one behaviour is best case (finding each child immediately, deleting it, then recreating on the end) while the other is worst case (not finding the child so searching to the end, with each iteration larger than the last). In profiling this divided loop proves to be much faster, when we destroy all the old objects first before re-instantiating, in all cases.

## Testing

Tested on Pixar's PointInstancedMedCity file.

Before: 
image

After:
image

Tested that initial import, re-import and refresh from USD no longer take a ridiculous time (>1minute) both in a cold and warm editor.

## Overall Product Risks

**Complexity:** low

**Halo Effect:** low

## Additional information

**Note to reviewers:**

**Reminder:**

### DIFF
--- a/package/com.unity.formats.usd/Runtime/Scripts/IO/Geometry/InstanceImporter.cs
+++ b/package/com.unity.formats.usd/Runtime/Scripts/IO/Geometry/InstanceImporter.cs
@@ -12,6 +12,7 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
+using System;
 using UnityEngine;
 using USD.NET;
 using PointInstancerSample = USD.NET.Unity.PointInstancerSample;
@@ -68,7 +69,6 @@ namespace Unity.Formats.USD
             SceneImportOptions options)
         {
             Matrix4x4[] transforms = sample.ComputeInstanceMatrices(scene, pointInstancerPath);
-            int i = 0;
 
             foreach (var protoRoot in sample.prototypes.targetPaths)
             {
@@ -98,6 +98,8 @@ namespace Unity.Formats.USD
             }
             */
 
+            Tuple<Matrix4x4, string, GameObject>[] instancesToCreate = new Tuple<Matrix4x4, string, GameObject>[sample.protoIndices.Length];
+            int i = 0;
             foreach (var index in sample.protoIndices)
             {
                 if (inactiveIds.Contains(index))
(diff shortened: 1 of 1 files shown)


package/com.unity.formats.usd/Runtime/Scripts/IO/Geometry/InstanceImporter.cs
```diff
@@ -98,6 +98,8 @@ public static void BuildSceneInstances(PrimMap primMap, SceneImportOptions optio
             }
             */
 
+            Tuple<Matrix4x4, string, GameObject>[] instancesToCreate = new Tuple<Matrix4x4, string, GameObject>[sample.protoIndices.Length];
+            int i = 0;
```
michaeljblain: Minor: we should call this something more descriptive e.g. instanceCount

michaeljblain (changes requested):
Alongside my minor suggestion I think it would be worth adding a comment for each loop describing what it is doing.

michaeljblain (approved):

package/com.unity.formats.usd/Runtime/Scripts/IO/Geometry/InstanceImporter.cs
```diff
@@ -68,14 +69,13 @@ public static void BuildSceneInstances(PrimMap primMap, SceneImportOptions optio
             SceneImportOptions options)
         {
             Matrix4x4[] transforms = sample.ComputeInstanceMatrices(scene, pointInstancerPath);
-            int i = 0;
 
             foreach (var protoRoot in sample.prototypes.targetPaths)
```
lee-aandrew: Does "proto" here mean prototype?
could we maybe change this to "prototype" and for below "prototypeRoot" if so?

package/com.unity.formats.usd/Runtime/Scripts/IO/Geometry/InstanceImporter.cs
```diff
@@ -117,36 +123,45 @@ public static void BuildSceneInstances(PrimMap primMap, SceneImportOptions optio
...
+                // The Transform hierarchy was created in the same order as it is now deleted,
+                // so this Find() operation is best-case performance.
                 var existingInstance = root.transform.Find(instanceName);
                 if (existingInstance != null)
                 {
                     GameObject.DestroyImmediate(existingInstance.gameObject);
                 }
 
-                var goInstance = GameObject.Instantiate(goMaster, root.transform);
+                instancesToCreate[instanceCount] = new Tuple<Matrix4x4, string, GameObject>(xf, instanceName, goMaster);
```
lee-aandrew: can we have the tuple here named?
so that we dont have to use Item1, Item2, Item3 when referencing them?

package/com.unity.formats.usd/Runtime/Scripts/IO/Geometry/InstanceImporter.cs
vickycl: Sure :)

package/com.unity.formats.usd/Runtime/Scripts/IO/Geometry/InstanceImporter.cs
vickycl: TIL about ValueTuples that can actually have named fields 😁 Done

PR #347, opened 2023-02-03
[USDU-304] Add performance Tests

lee-aandrew:
## Purpose of this PR

It would be beneficial to have Performance Tests for this package

Creating a new Unity Project to separate out existing functional tests and new performance tests using project setting compilation defines

**Ticket/Jira #:**

## Testing

## Overall Product Risks

**Complexity:**

**Halo Effect:**

## Additional information

**Note to reviewers:**

**Reminder:**


package/com.unity.formats.usd/package.json
```diff
@@ -2,7 +2,8 @@
     "dependencies": {
         "com.unity.timeline" : "1.0.0",
         "com.unity.modules.screencapture": "1.0.0",
-        "com.unity.modules.physics": "1.0.0"
+        "com.unity.modules.physics": "1.0.0",
+        "com.unity.test-framework.performance": "2.8.1-preview"
```
vickycl: This shouldn't be a dependency for the whole package, pls remove :)

TestProject/Usd-Development-Performance/ProjectSettings/ProjectVersion.txt
```diff
@@ -0,0 +1,2 @@
+m_EditorVersion: 2021.3.14f1
+m_EditorVersionWithRevision: 2021.3.14f1 (eee1884e7226)
```
vickycl: The ProjectVersion that is committed needs to match the lowest supported version so that it opens without downgrade warnings in that version, so 2019.4 (the Usd-development project has m_EditorVersion: 2019.4.26f1
m_EditorVersionWithRevision: 2019.4.26f1 (e0392c6b2363), so probably should match that)

package/com.unity.formats.usd/Tests/Performance/PerformanceBaseFixture.cs
```diff
@@ -0,0 +1,44 @@
...
+        public const int MeasurementCount = 3;
+        public const int IterationsPerMeasurement = 1;
+    }
+
+    public void Setup()
+    {
+        InitUsd.Initialize();
+        if (Directory.Exists(ArtifactsDirectoryFullPath))
+        {
+            TestUtilityFunction.DeleteFolder(ArtifactsDirectoryFullPath);
```
vickycl: You could call Cleanup() here instead :)

package/com.unity.formats.usd/Tests/Performance/PointInstancerPerformanceTests.cs
```diff
@@ -0,0 +1,113 @@
...
+            Larger
+        }
+
+        const string k_PointInstancerPrim10GUID = "ddba66229ec63364d8ce36112afcdbec";
+        const string k_PointInstancerPrim100GUID = "567d2c32dba4a364bb1fb80b134dc92d";
+        const string k_PointInstancerPrim1000GUID = "9ca154d52e7e03e4c989b66e33e08212";
+        // For Import as Prefab / Timeline it takes more than 3 minutes to run the test, so it wont be included there
+        const string k_PointInstancerPrim10000GUID = "b4f2b0c08e1cc0d43aa738154881fcfc";
+
+        private string DeterminePointInstancerSize(InstancerSize testSize)
```
vickycl: Something like `GetGUIDForPointInstancerSize` might be a more accurate name for what this function is doing?

.yamato/upm-ci.yml
```diff
@@ -8,10 +8,14 @@ test_platforms:
     type: Unity::VM
     image: package-ci/win10:stable
     flavor: b1.large
+    utr_install: curl -s https://artifactory.prd.it.unity3d.com/artifactory/unity-tools-local/utr-standalone/utr.bat --output utr.bat
+    utr_performance_test: utr --suite=editor --clean-library --category="Performance" --testproject=TestProject/Usd-Development-Performance --editor-location=.Editor --report-performance-data
```
vickycl: To report data to Observer I think you also need to pass a project ID, ie: `--performance-project-id=com.unity.formats.usd`

.yamato/upm-ci.yml
```diff
@@ -8,10 +8,14 @@ test_platforms:
...
   - name: mac
     type: Unity::VM::osx
     image: package-ci/mac:stable
     flavor: m1.mac
+    utr_install: curl -s https://artifactory.prd.it.unity3d.com/artifactory/unity-tools-local/utr-standalone/utr --output utr; chmod +x utr
+    utr_performance_test: ./utr --suite=editor --clean-library --category="Performance" --testproject=TestProject/Usd-Development-Performance --editor-location=.Editor --report-performance-data
```
vickycl: As above, `--performance-project-id=com.unity.formats.usd`

vickycl (changes requested):
I think this is a great start, nice PointInstancer tests :)

There's some basic functionality that I'd like to see tested in this first PR. Basic imports, exports, Reloads, and Force reimports with different data types would start to give us a solid baseline. In future also: use of some of the APIs; USDA, USDC, USDZ; a few other common data types, especially animation.

It would also be great if you could share what data initial runs of the tests has produced, as I'm curious whether the smallest tests are still long enough to be stable? I had a look on Observer but I can't see any results there, I think because of the missing project ID argument. Maybe fix that up and then check Observer is reporting the data :)

Some comments I can't do in line because GitHub doesn't support comments on empty/ binary files:
* Can we remove and add to the git ignore the VS generated .vs folder and .vsconfig file? 
* What is the .gitkeep, do we need it?
* The test project has a lot of additional packages that possibly aren't relevant. Looks like they might be the default template or something. Could we remove any that aren't necessary (and that won't just be pulled back in on opening, of course)?
* A lot of the project settings files look like just defaults, can they be .gitignore'd?
* Minor to the point of nit-picky: for the test project name I find Usd-Performance easier to parse than Usd-Development-Performance 😅

package/com.unity.formats.usd/package.json
lee-aandrew: idk how to write a reply for a standalone comment, so i will hijack this in-line comment and write it here 😅

- Your idea for additional tests sound good, I will work on adding those in as well! - thanks for the suggestions!
- I will try adding in the project ID argument, thanks for letting me know of this arg
- Will add the .vs folder and .vsconfig files to .gitignore
- .gitkeep is to explicitly keep empty folders tracked, in this case the Asset folder for the Performance test unity project
- I actually copied the package.json from our functional tests project, so if i can remove some unnecessary projects here, i will make the same changes there
- I'll look into the project settings - I think i just forgot to untrack them while experimenting with project specific Defines...
- I actually kind of want to rename our "Usd-Development" project to "Usd-Functionality" as well, which will make more sense with your suggestion of "Usd-Performance" - but for now, I will rename the new project to "Usd-Performance"!

package/com.unity.formats.usd/Tests/Performance/PointInstancerPerformanceTests.cs
lee-aandrew: Changed to GetGUIDForTestPointInstancerSize!

PR #350, opened 2023-03-03
Change our mac vm to mac 12 as the graphics requirement in our tests are failing on vms without metal graphics api

lee-aandrew:
## Purpose of this PR

**Ticket/Jira #:**
https://jira.unity3d.com/browse/USDU-318

## Testing

N/A
N/A

## Overall Product Risks

**Complexity:**
1

**Halo Effect:**
1

## Additional information
Slack thread:
https://unity.slack.com/archives/C0JHA6J4C/p1677792279052949

This change is required for our tests to run on the Trunk version of Unity, older versions of Unity did not require this change
Therefore it does unnecessarily change the minimum requirements for our tests on older versions of Unity

**Reminder:**

### DIFF
--- a/.yamato/upm-ci.yml
+++ b/.yamato/upm-ci.yml
@@ -10,8 +10,8 @@ test_platforms:
     flavor: b1.large
   - name: mac
     type: Unity::VM::osx
-    image: package-ci/macos-12:stable
-    flavor: m1.mac
+    image: slough-ops/macos-11-base:stable
+    flavor: b1.large
 test_backends:
   - name: il2cpp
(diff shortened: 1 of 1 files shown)


vickycl (approved):
As discussed offline, I think we should switch back to mac:stable if we can if the issue gets resolved in future. We should be testing with the minimum spec for older versions of Unity even if MacOS < 11 is not feasible for our trunk testing. However in the interest of getting the testing fixed let's go with this for now.

lee-aandrew:
Thanks for the review!

PR #358, opened 2023-03-21
[USDU-329] Ignores two unstable test cases until a fix has been found

lee-aandrew:
## Purpose of this PR

**Ticket/Jira #:**

[[USDU-329]](https://jira.unity3d.com/browse/USDU-329)

Test cases:
 - ChangingUsdPayloadStateFromUnloadedToLoaded_MarksSceneDirty
 - ChangingUsdPayloadStateFromLoadedToUnloaded_MarksSceneDirty
are randomly failing from time to time

 For now set them to ignore until we can find a proper fix

## Testing

Sets two test cases that was stable and passing as Ignore as they are no longer producing consistent passing results

None

## Overall Product Risks

The actual functionality of the feature that the two test cases are testing may have become inconsistent
But we need more time to investigate this, and for now, this PR will set the two test cases as Ignored

**Complexity:**

Min.

**Halo Effect:**

## Additional information

**Note to reviewers:**

Will work on why these two cases have become inconsistent in the near future

**Reminder:**

### DIFF
--- a/package/com.unity.formats.usd/Tests/Editor/USDPayloadComponentTests.cs
+++ b/package/com.unity.formats.usd/Tests/Editor/USDPayloadComponentTests.cs
@@ -207,6 +207,7 @@ namespace Unity.Formats.USD.Tests
             Assume.That(usdPayload.IsLoaded, Is.False, "UsdPayload.IsLoaded should be set to false.");
         }
 
+        [Ignore("[USDU-329] Unstable test result")]
         [UnityTest]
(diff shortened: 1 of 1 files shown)


vickycl (changes requested):
I'd rather not disable tests without a clear plan to get them fixed, especially as these are brand new tests of a recent bug fix. Please add more details to the Jira ticket so that we can ensure this is looked into and not lost :)

vickycl (approved):
Thanks for updating the Jira :)

vickycl:
Merging as investigation is underway and the unstable test is blocking all other PRs :)

PR #359, opened 2023-03-29
[USDU-328] Fix for IndexOutOfRangeException when importing UsdSkel

vickycl:
## Purpose of this PR

**Ticket/Jira #:** USDU-328 / #352 
Fixes the IndexOutOfRangeException when importing Pixar's Human Female UsdSkel example. My PR #343 introduced a regression for Skinned Meshes, which use the cached triangulatedFaceVertexIndices to map from the new unwelded vertices to the old ones. My change to switch from a List to an array hid that there was no longer a copy, so the triangulatedFaceVertexIndices was now referencing the same array as faceVertexIndices, which was later changed. I've changed all the arrays that are needed to hold backups in this code to have a copy.

There was also an unreported issue with ALab 2 where some of the arbitrary primvars are stated to have face varying interpolation, but do not have one value per vertex per face, which is a (potentially problematic) requirement in our import process. I changed this to give a warning and use GuessInterpolation to find the next best behaviour. 

This PR also optimizes ImportSkinning a bit, by removing costly gets and sets to native arrays, and instead working on a managed array that is fed into a native array later.

## Testing

Manually tested on Pixar's human female, ALab and another triangle-heavy file. I'd like to add some automated tests but this isn't straightforward for this edge case (we have existing tests in this area that didn't catch it).

The copying of the array does unfortunately regress the performance. I minimized as much as possible by using the lightweight Buffer.BlockCopy, but in the 'SphereCube' sample this increases the Triangulate time by ~40%. However, coupled with the optimization for TriangulateAttributes(), overall the import for ALab improved from 45s to 39s on my machine, so the extra <1s in BackupTopology() is somewhat less noticeable.

## Overall Product Risks

**Complexity:** Low

**Halo Effect:** Low

## Additional information

**Note to reviewers:**

**Reminder:**

### DIFF
--- a/package/com.unity.formats.usd/Runtime/Scripts/IO/Geometry/MeshImporter.cs
+++ b/package/com.unity.formats.usd/Runtime/Scripts/IO/Geometry/MeshImporter.cs
@@ -361,13 +361,17 @@ namespace Unity.Formats.USD
 
             int isNotConstant = weightsInterpolation.GetString() == UsdGeomTokens.constant ? 0 : 1;
 
+            // Convert the USD joint weights and indices into a Unity BoneWeights array, including remapping to the new triangulated vertices where required
             for (var vertexIndex = 0; vertexIndex < vertexCount; vertexIndex++)
             {
+                // For every new vertex, get the original vertex if it has been triangulated
                 var originalVertexIndex = remapIndices ? usdMesh.triangulatedFaceVertexIndices[vertexIndex] * weightsElementSize : vertexIndex * weightsElementSize;
                 var vertexLookupIndex = isNotConstant * originalVertexIndex;
 
+                // This intermediary array is faster than getting/setting to a NativeArray in a loop
                 bonesPerVertexArray[vertexIndex] = (byte)weightsElementSize;
 
+                // Create a BoneWeight for every element at every vertex
                 for (var weightElementIndex = 0; weightElementIndex < weightsElementSize; weightElementIndex++)
                 {
                     var boneWeight = new BoneWeight1();
--- a/package/com.unity.formats.usd/Runtime/Scripts/IO/Geometry/MeshImporter.cs
+++ b/package/com.unity.formats.usd/Runtime/Scripts/IO/Geometry/MeshImporter.cs
@@ -346,8 +346,9 @@ namespace Unity.Formats.USD
                 throw new Exception("Joints weights information are invalid or empty for: " + path);
             }
 
-            var bonesPerVertex = new NativeArray<byte>(unityMesh.vertexCount, Allocator.Temp);
-            var boneWeights1 = new NativeArray<BoneWeight1>(unityMesh.vertexCount * weightsElementSize, Allocator.Temp);
+            int vertexCount = unityMesh.vertexCount;
+            var bonesPerVertexArray = new byte[vertexCount];
+            var boneWeightsArray = new BoneWeight1[vertexCount * weightsElementSize];
 
             // Remap the vertex indices if mesh attribute have been converted to faceVarying
             var remapIndices = weightsInterpolation.GetString() == UsdGeomTokens.vertex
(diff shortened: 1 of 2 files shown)


package/com.unity.formats.usd/Runtime/Scripts/IO/Geometry/MeshImporter.cs
```diff
@@ -346,31 +346,39 @@ public static void ImportSkinning(string path, SanitizedMeshSample usdMesh, Mesh
...
+            var boneWeightsArray = new BoneWeight1[vertexCount * weightsElementSize];
 
             // Remap the vertex indices if mesh attribute have been converted to faceVarying
             var remapIndices = weightsInterpolation.GetString() == UsdGeomTokens.vertex
                 && usdMesh.arePrimvarsFaceVarying;
 
             int isNotConstant = weightsInterpolation.GetString() == UsdGeomTokens.constant ? 0 : 1;
 
-            for (var i = 0; i < unityMesh.vertexCount; i++)
+            for (var vertexIndex = 0; vertexIndex < vertexCount; vertexIndex++)
```
michaeljblain, line 365: We should add a comment describing what this loop is doing.

package/com.unity.formats.usd/Runtime/Scripts/IO/Geometry/MeshImporter.cs
```diff
@@ -346,31 +346,39 @@ public static void ImportSkinning(string path, SanitizedMeshSample usdMesh, Mesh
...
             {
-                var unityIndex = remapIndices ? usdMesh.triangulatedFaceVertexIndices[i] * weightsElementSize : i * weightsElementSize;
-                var usdIndex = isNotConstant * unityIndex;
+                var originalVertexIndex = remapIndices ? usdMesh.triangulatedFaceVertexIndices[vertexIndex] * weightsElementSize : vertexIndex * weightsElementSize;
+                var vertexLookupIndex = isNotConstant * originalVertexIndex;
 
-                bonesPerVertex[i] = (byte)weightsElementSize;
+                bonesPerVertexArray[vertexIndex] = (byte)weightsElementSize;
```
michaeljblain, line 372: `bonesPerVertexArray` is just an array with all elements having the same constant value. Can we not just initialize the NativeArray directly with all values at the end of this method and thus save the intermediate allocation and loop assignment?

michaeljblain (changes requested):
Looks good. Some minor feedback.

package/com.unity.formats.usd/Runtime/Scripts/IO/Geometry/MeshImporter.cs
vickycl: I couldn't find a NativeArray constructor that took a single value to initialize to, and the getter and setter for NativeArray are slow so it was measurably faster to allocate the intermediary C# array and construct the native one from it than assigning each element during the loop. I tried doing it in a separate loop for better cache coherency but any performance difference was indistinguishable from noise. I haven't been able to measure the GC impact of the additional array, though.

package/com.unity.formats.usd/Runtime/Scripts/IO/Geometry/MeshImporter.cs
vickycl: Done :)

package/com.unity.formats.usd/Runtime/Scripts/IO/Geometry/MeshImporter.cs
michaeljblain: Thanks for investigating. Let's move forward with what we have so as to not hold things up.

michaeljblain (approved):

package/com.unity.formats.usd/Runtime/Scripts/IO/SanitizedSamples.cs
```diff
@@ -486,44 +492,67 @@ static void FlattenPrimvar(Primvar<object> primvar)
...
         }
 
-        internal static void TriangulateAttributes<T>(ref T[] values, int[] faceVertexCount, bool changeHandedness)
+        internal static void TriangulateAttributes<T>(ref T[] values, int[] faceVertexCounts, bool changeHandedness)
         {
-            var newValues = new List<T>();
-            var last = 0;
-            for (var i = 0; i < faceVertexCount.Length; i++)
+            int newFacesCount = 0;
+            for (int i = 0; i < faceVertexCounts.Length; i++)
```
lee-aandrew: a little nitpick but can we perhaps change this to a foreach loop for better readability?

package/com.unity.formats.usd/Runtime/Scripts/IO/SanitizedSamples.cs
```diff
@@ -486,44 +492,67 @@ static void FlattenPrimvar(Primvar<object> primvar)
...
-        /// <remarks> If the interpolation of the array to convert is not known, it will be guessed based on the length of the array. </remarks
+        /// <remarks> If the interpolation of the array to convert is not known, it will be guessed based on the length of the array. </remarks>
         void ConvertInterpolationToFaceVarying<T>(ref T[] values, int[] vertexIndices, bool changeHandedness = false, TfToken interpolation = null)
         {
             if (values == null)
                 return;
 
-            if (interpolation == null)
-                interpolation = GuessInterpolation(values.Length);
+            var newInterpolation = GuessInterpolation(values.Length);
```
lee-aandrew, line 547: it seems like we now call GuessInterpolation no matter what the given interpolation was - i am guessing that when interpolation is given it should in most cases match the result of GuessInterpolation? 

Is this done as validation?
if so, is GuessInterpolation a quick process? If it is, should we now provide a "Skip Interpolation Validation" flag for more experienced users?  (Assuming this is shared code when Updating a USD file under the "Conversions" setting)

package/com.unity.formats.usd/Runtime/Scripts/IO/SanitizedSamples.cs
vickycl: I guess that's a matter of preference, personally I find a for loop more readable 😅 Maybe that's just me though, if you prefer foreach I will change it

package/com.unity.formats.usd/Runtime/Scripts/IO/SanitizedSamples.cs
vickycl: Yep, that's the intention. It turns out now that we have arbitrary primvars we can't trust the authoring DCC of the USD file to actually have set the correct interpolation token. In the majority of cases no interpolation is passed in, so it was always performing the 'guess' anyway. GuessInterpolation is pretty cheap and this won't be a particularly hot path so I don't think we need to expose it to users. This whole area is quite hacky because we're forcing all the attributes to be face varying instead of just using what they're authored as.

package/com.unity.formats.usd/Runtime/Scripts/IO/SanitizedSamples.cs
lee-aandrew: thank you, but as you said its a matter of preference so its okay, i also see that you used for loops instead of foreach for the loops below, so i think its better to be consistent

Plus I once read that foreach loop was faster than for-loops, but after checking it again, it seems that i read it wrong and there's no real difference in real-world usage

PR #360, opened 2023-03-29
[USDU-329] Change the wait criteria of flaky test cases

lee-aandrew:
## Purpose of this PR

This PR attempts to circumvent the current error we are having with Unity ver. 2020.3 where 2 of our test cases, 
- ChangingUsdPayloadStateFromUnloadedToLoaded_MarksSceneDirty
- ChangingUsdPayloadStateFromLoadedToUnloaded_MarksSceneDirty

are often failing due to not waiting until the Payload Load / Unload has not fully updated the root payload GameObject

**Ticket/Jira #:**
[USDU-329](https://jira.unity3d.com/browse/USDU-329)

## Testing
Changes the testcases:
- ChangingUsdPayloadStateFromUnloadedToLoaded_MarksSceneDirty
- ChangingUsdPayloadStateFromLoadedToUnloaded_MarksSceneDirty

Instead of doing an Assert on scene.isDirty, we wait until the scene has been dirtied, and then do an assert on the rootPayload Object's child count

This should solve the problem of not waiting long enough as well as scoping more into which GameObject within the scene has been changed

N/A

## Overall Product Risks

**Complexity:**

1

**Halo Effect:**

## Additional information

**Note to reviewers:**

I at first did not want to change the test cases to this extent, and wanted to gain more information on why only on UPM we are getting this problem with waiting for proper update
[Slack Thread 1](https://unity.slack.com/archives/C0JHA6J4C/p1680116747100079) & [Slack Thread 2](https://unity.slack.com/archives/C02A3U0UP9S/p1680115238408449)

Though I do believe the changes here are good for these 2 test cases, i would like to learn more about this UPM and Wait issue on Unity 2020.3 so that this doesnt happen again in the future - will update as soon as i get more info on this

**Reminder:**

### DIFF
--- a/package/com.unity.formats.usd/Tests/Editor/USDPayloadComponentTests.cs
+++ b/package/com.unity.formats.usd/Tests/Editor/USDPayloadComponentTests.cs
@@ -207,6 +207,7 @@ namespace Unity.Formats.USD.Tests
             Assume.That(usdPayload.IsLoaded, Is.False, "UsdPayload.IsLoaded should be set to false.");
         }
 
+        [Ignore("[USDU-329] Unstable test result")]
         [UnityTest]
         public IEnumerator ChangingUsdPayloadStateFromUnloadedToLoaded_MarksSceneDirty()
         {
@@ -224,6 +225,7 @@ namespace Unity.Formats.USD.Tests
             Assert.IsTrue(m_UnityScene.isDirty, "Scene should be dirty after changing UsdPayload objects");
         }
 
+        [Ignore("[USDU-329] Unstable test result")]
         [UnityTest]
         public IEnumerator ChangingUsdPayloadStateFromLoadedToUnloaded_MarksSceneDirty()
         {
--- a/package/com.unity.formats.usd/Tests/Editor/USDPayloadComponentTests.cs
+++ b/package/com.unity.formats.usd/Tests/Editor/USDPayloadComponentTests.cs
@@ -208,6 +208,31 @@ namespace Unity.Formats.USD.Tests
         }
 
         [UnityTest]
+        public IEnumerator ChangingUsdPayloadStateFromUnloadedToLoaded_MarksSceneDirty()
+        {
+            const int timeOutSeconds = 3;
+
+            m_usdAsset.m_payloadPolicy = PayloadPolicy.DontLoadPayloads;
+            m_usdAsset.Reload(forceRebuild: true);
+            yield return null;
+
+            EditorSceneManager.SaveScene(m_UnityScene, GetUnityScenePath());
+            Assert.IsFalse(m_UnityScene.isDirty, "Scene should not be dirty after after saving");
+
+            float loadingTime = 0;
+            m_usdRoot.GetComponentInChildren<UsdPayload>().Load();
+
+            while (!m_UnityScene.isDirty && loadingTime < timeOutSeconds)
+            {
+                loadingTime += Time.deltaTime;
+                yield return null;
+            }
+
+            Assert.Less(loadingTime, timeOutSeconds);
+            Assert.True(m_UnityScene.isDirty);
+        }
+
+        [UnityTest]
         public IEnumerator ChangingUsdPayloadStateFromUnloadedToLoaded_GeneratesPayloadObjects()
         {
             m_usdAsset.m_payloadPolicy = PayloadPolicy.DontLoadPayloads;
@@ -217,7 +242,6 @@ namespace Unity.Formats.USD.Tests
             EditorSceneManager.SaveScene(m_UnityScene, GetUnityScenePath());
             Assert.IsFalse(m_UnityScene.isDirty, "Scene should not be dirty after after saving");
 
-            // Change payload state and wait one frame to ensure Update runs
             var payloadRoot = m_usdRoot.GetComponentInChildren<UsdPayload>();
             var payloadRootChildCount = payloadRoot.transform.childCount;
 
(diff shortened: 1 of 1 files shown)


package/com.unity.formats.usd/Tests/Editor/USDPayloadComponentTests.cs
```diff
@@ -218,14 +218,20 @@ public IEnumerator ChangingUsdPayloadStateFromUnloadedToLoaded_MarksSceneDirty()
             Assert.IsFalse(m_UnityScene.isDirty, "Scene should not be dirty after after saving");
 
             // Change payload state and wait one frame to ensure Update runs
-            m_usdRoot.GetComponentInChildren<UsdPayload>().Load();
-            yield return null;
+            var payloadRoot = m_usdRoot.GetComponentInChildren<UsdPayload>();
+            var payloadRootChildCount = payloadRoot.transform.childCount;
+
+            payloadRoot.Load();
+            yield return new WaitUntil(() => m_UnityScene.isDirty == true);
```
vickycl: The original test was to address a bug where the scene was not set to dirty at all after changing the payload setting. In that case, this modification to the test would run forever (or presumably until a UTR timeout, which would mean no useful information from the failure). 

However it looks like we can't rely on UsdPayload APIs to correctly report their own status either (`UsdPayload.IsLoaded()` will return true _before_ it's actually been loaded, seems weird to me and should be reviewed) so we should find a more stable way to wait on whether the payload change has been registered before we check whether the scene is dirty. (and also probably change the test name back to `_MarksSceneDirty`, because that's the behaviour the test was written for)

package/com.unity.formats.usd/Tests/Editor/USDPayloadComponentTests.cs
```diff
@@ -218,14 +218,20 @@ public IEnumerator ChangingUsdPayloadStateFromUnloadedToLoaded_MarksSceneDirty()
...
+
+            Assert.IsTrue(payloadRoot == null, "Payload Load should reset USD related components and older references to it should be null");
 
-            Assert.IsTrue(m_UnityScene.isDirty, "Scene should be dirty after changing UsdPayload objects");
+            var newPayloadRoot = m_usdRoot.GetComponentInChildren<UsdPayload>();
+            Assert.AreNotEqual(payloadRootChildCount, newPayloadRoot.transform.childCount, "Payload Load should generate the payload objects");
```
vickycl: I think the addition of checking that the payload load has actually changed the hierarchy is good, as I think that is behaviour we don't check (at least that I can see in this file!). Could you please break that out into a separate test- as long as there isn't already a test doing that in a different file :) 

I believe the inputs to the Assert need to be switched, as the 'expected' to be compared against would be the newPayloadRoot, with the 'actual' being the payloadRootChildCount. Small thing, but makes diagnosing a test failure easier.

package/com.unity.formats.usd/Tests/Editor/USDPayloadComponentTests.cs
```diff
@@ -235,10 +241,13 @@ public IEnumerator ChangingUsdPayloadStateFromLoadedToUnloaded_MarksSceneDirty()
             Assert.IsFalse(m_UnityScene.isDirty, "Scene should not be dirty after after saving");
 
             // Change payload state and wait one frame to ensure Update runs
-            m_usdRoot.GetComponentInChildren<UsdPayload>().Unload();
-            yield return null;
+            var payloadRoot = m_usdRoot.GetComponentInChildren<UsdPayload>();
+            var payloadRootChildCount = payloadRoot.transform.childCount;
+
+            payloadRoot.Unload();
+            yield return new WaitUntil(() => m_UnityScene.isDirty == true);
```
vickycl: As above :)

vickycl (changes requested):
Well spotted on the cause of this instability- it could be the ordering of calling different Update methods at fault. 

The changes have however changed the condition that it is testing for, which I think is a bad idea. We still need to check that the scene is actually marked dirty, as that is the bug that this test was written for. I've added comments in line, but ping me if you want to discuss :)

package/com.unity.formats.usd/Tests/Editor/USDPayloadComponentTests.cs
lee-aandrew: I decided to keep the `Generate` / `Destroys` tests and bring back the 2 dirties tests as I agree with your point as well,

But I didnt really a "clean" way to test this, so I decided to change it to a time out, and make sure that the scene is marked within 3 seconds - I wish Unity's `WaitUntil` had a way add a timeout value, but sadly, that doesnt seem to be the case - let me know what you think about this method

package/com.unity.formats.usd/Tests/Editor/USDPayloadComponentTests.cs
```diff
@@ -210,35 +210,92 @@ public IEnumerator IsLoaded_IsFalse_WhenIsUnloadedFromComponent()
...
             yield return null;
 
             EditorSceneManager.SaveScene(m_UnityScene, GetUnityScenePath());
             Assert.IsFalse(m_UnityScene.isDirty, "Scene should not be dirty after after saving");
 
-            // Change payload state and wait one frame to ensure Update runs
+            float loadingTime = 0;
             m_usdRoot.GetComponentInChildren<UsdPayload>().Load();
+
+            while (!m_UnityScene.isDirty && loadingTime < timeOutSeconds)
```
vickycl: Can we make it not a timeout but a number of frames? A timeout risks becoming unstable in the future if the frame rate becomes slower, for example. Can we increment a counter in the loop and after three frames fail?

Also please add a comment above to explain eg 'Wait at least one frame to ensure Update runs, break out if we hit either the condition or have waited for too many frames'.

package/com.unity.formats.usd/Tests/Editor/USDPayloadComponentTests.cs
```diff
@@ -210,35 +210,92 @@ public IEnumerator IsLoaded_IsFalse_WhenIsUnloadedFromComponent()
...
+            {
+                loadingTime += Time.deltaTime;
+                yield return null;
+            }
+
+            Assert.Less(loadingTime, timeOutSeconds);
+            Assert.True(m_UnityScene.isDirty);
```
vickycl: Please add back in the test message, and also one for the case where we breakout

package/com.unity.formats.usd/Tests/Editor/USDPayloadComponentTests.cs
```diff
@@ -210,35 +210,92 @@ public IEnumerator IsLoaded_IsFalse_WhenIsUnloadedFromComponent()
...
+        [UnityTest]
+        public IEnumerator ChangingUsdPayloadStateFromUnloadedToLoaded_GeneratesPayloadObjects()
+        {
+            m_usdAsset.m_payloadPolicy = PayloadPolicy.DontLoadPayloads;
+            m_usdAsset.Reload(forceRebuild: true);
             yield return null;
 
-            Assert.IsTrue(m_UnityScene.isDirty, "Scene should be dirty after changing UsdPayload objects");
+            EditorSceneManager.SaveScene(m_UnityScene, GetUnityScenePath());
+            Assert.IsFalse(m_UnityScene.isDirty, "Scene should not be dirty after after saving");
```
vickycl: I know this was already there, but as you're in this code could you remove the extra "after" here please? :)

package/com.unity.formats.usd/Tests/Editor/USDPayloadComponentTests.cs
vickycl: I think let's wait on a number of frames instead of a time- 3 seconds is a long time for a test to run even if it is failing, and a shorter test time risks becoming unstable in the future if Unity execution gets slower. The expected behaviour is that the payload should be loaded in the next frame, so waiting 3 or 4 frames should be enough to confirm this, and if not then we have bigger issues :D

vickycl: I think we should add the timeout/ breakout in these other two tests too- in the (hopefully unlikely) scenario that this breaks in future and the tests are run in a different order, these will get stuck in the loop of waiting for the bool to change and not diagnose the actual issue which the other tests would flag.

vickycl (changes requested):

vickycl (approved):
Thanks of the iterations! Look good 👍

lee-aandrew:
Thanks for the reviews! @vickycl

PR #364, opened 2023-03-31
[USDU-342] Fix for importing materials exported with a newer version of USD.

vickycl:
## Purpose of this PR

**Ticket/Jira #:** USDU-342 / #353 

In USD 21.11, the type of the UsdPrimvarReader.varname was changed from token to string. This change now handles both the token and string case in the material importer.
We should work on adding some test cases of USD files exported from newer USD versions to our suite of test files.

## Testing

Tested files exported in both the older and newer versions to check both were as expected. 

Unlikely, simple change.

## Overall Product Risks

**Complexity:** 1

**Halo Effect:** 1

## Additional information

**Note to reviewers:**

**Reminder:**

### DIFF
--- a/package/com.unity.formats.usd/CHANGELOG.md
+++ b/package/com.unity.formats.usd/CHANGELOG.md
@@ -2,6 +2,8 @@
 
 ## Unreleased
 ### Bug Fixes
+- Fixed a bug where importing materials exported from a newer USD version would fail.
+
 - Fixed an import bug causing instanced primitives not to be sanitized.
 - Disabled plugins on unsupported platforms.
 - Fixed a bug causing PointInstances to be duplicated.
--- a/package/com.unity.formats.usd/CHANGELOG.md
+++ b/package/com.unity.formats.usd/CHANGELOG.md
@@ -2,7 +2,7 @@
 
 ## Unreleased
 ### Bug Fixes
-- Fixed a bug where importing materials exported from a newer USD version would fail.
+- Fixed a bug where importing materials exported from USD version >= 21.11 would fail.
 
 - Fixed an import bug causing instanced primitives not to be sanitized.
 - Disabled plugins on unsupported platforms.
--- a/package/com.unity.formats.usd/Runtime/Scripts/IO/Materials/MaterialImporter.cs
+++ b/package/com.unity.formats.usd/Runtime/Scripts/IO/Materials/MaterialImporter.cs
@@ -242,6 +242,11 @@ namespace Unity.Formats.USD
                                 uvPrimvar = pxr.UsdCs.VtValueToTfToken(value).ToString();
                             else
                                 uvPrimvar = value;
+
+                            if (string.IsNullOrEmpty(uvPrimvar))
+                            {
+                                Debug.LogWarning($"uvPrimvar at <{connPath}> was unable to be read from USD file.");
+                            }
                         }
                         else
                         {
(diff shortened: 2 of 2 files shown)


package/com.unity.formats.usd/Runtime/Scripts/IO/Materials/MaterialImporter.cs
```diff
@@ -236,7 +236,17 @@ System.Collections.Generic.List<string> primvars
                         if (attr != null)
                         {
                             var value = attr.Get(scene.Time);
-                            uvPrimvar = pxr.UsdCs.VtValueToTfToken(value).ToString();
+
+                            // This value type can be a string or a token, depending on USD version
+                            if (string.IsNullOrEmpty(value))
+                                uvPrimvar = pxr.UsdCs.VtValueToTfToken(value).ToString();
```
lee-aandrew, line 245: One question here, it seems like if `value` is null or is an empty string, we set set
`uvPrimvar` to `pxr.UsdCs.VtValueToTfToken(null).ToString();` or `pxr.UsdCs.VtValueToTfToken("").ToString();`

would `uvPrimvar` be set to some sort of a default in this case, or just straight go to line 248?

vickycl: `value` is a VtValue, not a string. We currently have two cases in this code depending which USD version we're on. In one case, VtValue contains string data, so can be cast to a string and String.IsNullOrEmpty will be false -> so treat as a string. In the other case, where it's a token, it can't be cast to a string so String.IsNullOrEmpty returns true and we can use VtValueToTfToken to extract the token from it.

vickycl: @lee-aandrew I was reluctant to add another API call into Pxr code, but as this isn't a hot path, I've refactored it for the sake of clarity. Hope this makes more sense now?

lee-aandrew: oh cool thanks I didnt know String.IsNullOrEmpty would return true for objects that cant be casted to a string type
And double thanks for the refactor!

lee-aandrew (approved):
LGTM!

package/com.unity.formats.usd/CHANGELOG.md
```diff
@@ -2,6 +2,8 @@
 
 ## Unreleased
 ### Bug Fixes
+- Fixed a bug where importing materials exported from a newer USD version would fail.
```
michaeljblain: I'd modify this to be more explicit i.e. versions >= 21.11

package/com.unity.formats.usd/Runtime/Scripts/IO/Materials/MaterialImporter.cs
```diff
@@ -235,8 +235,16 @@ System.Collections.Generic.List<string> primvars
                         var attr = scene.GetAttributeAtPath(connPath);
                         if (attr != null)
                         {
-                            var value = attr.Get(scene.Time);
-                            uvPrimvar = pxr.UsdCs.VtValueToTfToken(value).ToString();
+                            pxr.VtValue value = attr.Get(scene.Time);
+
+                            // This value type can be a string or a token, depending on USD version
```
michaeljblain: I'd also add USD version >= 21.11 as part of this comment.

michaeljblain (changes requested):
Some minor comments. Otherwise looks good.

package/com.unity.formats.usd/CHANGELOG.md
vickycl: Thanks for confirming, I was being vague because I wasn't sure exactly which version introduced the change 😅

michaeljblain (approved):
Looks good - thanks for the changes.

PR #365, opened 2023-03-31
[USDU-305][USDU-345] Convert Imported PointInstancer transforms to Unity transform space & Adds Testcase for Importing Instance USD Files and their generated GameObject Axis direction

lee-aandrew:
[[USDU-345]](https://jira.unity3d.com/browse/USDU-345?filter=-1)
[[USDU-305]](https://jira.unity3d.com/browse/USDU-305)

## Purpose of this PR

**Ticket/Jira #:**

Unity is of Left handed orientation, while many DCCs are of Right handed orientation
We need a test case to test for correct Z-Axis placement when Importing a instancer USD file
This PR also includes the fix that Jeremy has suggested in [[USDU-305]](https://jira.unity3d.com/browse/USDU-305)

## Testing

I created a simple USD file that generates a model that has 8 unique vertex positions
The test imports this model, and runs through all its vertex points
if the imported mesh's vertex point is not included in the expected vertex points (ones defined in the usd file) it will fail the test
Do note we expect the vertex point to have its Z-Axis value to be flipped to compensate for the Z-Axis difference between Unity and DCCs - What is of 'positive' value on a DCC like Blender will be turned negative in Unity, but this is to import the object as-is from the DCC

N/A

## Overall Product Risks

N/A

**Complexity:**

Min

**Halo Effect:**

## Additional information

**Note to reviewers:**

**Reminder:**


package/com.unity.formats.usd/Tests/USD.NET.Unity/InstancingTests.cs
```diff
@@ -94,5 +98,21 @@ public void PointInstancerTest()
             }
             scene.Close();
         }
+
+        [Test]
+        [Ignore("TODO: Enable this when fix for this has been applied")]
```
vickycl: Can you add the Bug number please?

package/com.unity.formats.usd/Tests/Common/Data/UsdImportTests/UsdInstanceVertexTest.usda
```diff
@@ -0,0 +1,42 @@
...
+{
+    def PointInstancer "instancers" (
+        kind = "group"
+    )
+    {
+        int64[] invisibleIds = []
+        point3f[] positions = [(0, 0, 0)]
+        int[] protoIndices = [0]
+        rel prototypes = </zAxisTest_Instanced/instancers/uniquePointsMesh>
+        uniform token orientation = "leftHanded"
```
vickycl: Can we have additional test cases that tests what happens when this is rightHanded and when the upAxis is different?

package/com.unity.formats.usd/Tests/Common/Data/UsdImportTests/UsdInstanceVertexTest.usda
```diff
@@ -0,0 +1,42 @@
+#usda 1.0
+(
+    doc = "Blender v3.3.1"
+    metersPerUnit = 1
+    upAxis = "Y"
+)
+
+def Xform "zAxisTest_Instanced"
```
vickycl: For the sake of reusability, I'd recommend calling this something generic rather than specifically about the test case. This is just a USD file containing a point instancer, and what we test with it is up to the test code. You could specify the orientation setting maybe, to distinguish it from other files.

package/com.unity.formats.usd/Tests/USD.NET.Unity/InstancingTests.cs
```diff
@@ -94,5 +98,21 @@ public void PointInstancerTest()
...
+        {
+            const string testFileName = "UsdInstanceVertexTest";
+            HashSet<Vector3> expectedVertexPoints = new HashSet<Vector3> { new Vector3(-1, 0, 3), new Vector3(-1, 0, 0), new Vector3(0, -1, 1), new Vector3(-2, -3, -1), new Vector3(-2, -1, 1), new Vector3(-3, -0.5f, -1), new Vector3(-1, -2, 1), new Vector3(-5, -2, -1) };
+            var testScene = ImportHelpers.InitForOpen(GetTestAssetPath(testFileName));
+
+            var testInstanceObjectMesh = ImportHelpers.ImportSceneAsGameObject(testScene).GetComponentInChildren<MeshFilter>().mesh;
+
+            foreach (var vertex in testInstanceObjectMesh.vertices)
+            {
+                Assert.That(expectedVertexPoints.Contains(vertex), $"Vertex <{vertex}> is not found in \n\t{string.Join("\n\t", expectedVertexPoints)}");
```
vickycl: I think this test could fail just because of floating point errors transforming from one axis to another? Also the LINQ inside the foreach here feels overkill. The points, I think, should be in the same order still, so just a for loop comparing the points at each index (with a delta) should be enough

vickycl (changes requested):
This test iteration is much more readable :)

package/com.unity.formats.usd/Tests/USD.NET.Unity/InstancingTests.cs
lee-aandrew: In Unity, the imported file's mesh.vertices seem to be in a different order then what we specify in the usda file so I dont think checking each index would be valid, unless we figure out how the mesh.vertices are ordered

When you mean "transforming from one axis to another" do you mean the specified X value becoming the Z value?

vickycl: I guess because the verts are unwelded there will be duplicate verts in the Unity mesh than the USD mesh, although I'm surprised they're not in the same order. Do you happen to have recorded what the new order was? :)

Yeah basically, when we do the matrix multiplication to transform from one to the other we could introduce slight variations in the floats

vickycl: I just came across a test in another file: `MeshInstancesLoading_MeshesSanitized_True` - can any of this logic be shared?

lee-aandrew: i'll check that out! thanks!

lee-aandrew: > I just came across a test in another file: `MeshInstancesLoading_MeshesSanitized_True` - can any of this logic be shared?

Unfortunately this doesnt work :(
The vertices on Unity side is re-ordered and unwelded for our test asset

package/com.unity.formats.usd/Dependencies/USD.NET.Unity/Geometry/PointInstancerSample.cs
```diff
@@ -48,16 +48,18 @@ public Matrix4x4[] ComputeInstanceMatrices(Scene scene, string primPath)
             var prim = scene.GetPrimAtPath(primPath);
             var pi = new pxr.UsdGeomPointInstancer(prim);
             var xforms = new pxr.VtMatrix4dArray();
+            var timeCode = scene.Time == null ? pxr.UsdTimeCode.Default() : scene.Time;
 
             // For the base time from Pixar docs:
             // "If your application does not care about off-sample interpolation, it can supply the same value for baseTime that it does for time."
-            pi.ComputeInstanceTransformsAtTime(xforms, scene.Time == null ? pxr.UsdTimeCode.Default() : scene.Time, scene.Time == null ? pxr.UsdTimeCode.Default() : scene.Time);
```
vickycl, line 54: nice refactor :)

package/com.unity.formats.usd/Tests/USD.NET.Unity/InstancingTests.cs
```diff
@@ -1,7 +1,11 @@
 using System;
+using System.Collections.Generic;
+using System.IO;
```
vickycl: I don't think these are necessary anymore with the latest refactor of the test code :)

package/com.unity.formats.usd/Tests/USD.NET.Unity/InstancingTests.cs
```diff
@@ -94,5 +98,34 @@ public void PointInstancerTest()
...
+                new Vector3(-5, -2, -1)
+            };
+            var testScene = ImportHelpers.InitForOpen(GetTestAssetPath(testFileName));
+
+            var testInstanceObjectMesh = ImportHelpers.ImportSceneAsGameObject(testScene).GetComponentInChildren<MeshFilter>().mesh;
+
+            foreach (var vertex in testInstanceObjectMesh.vertices)
+            {
+                // Flip the z-axis value to compensate for the z-axis difference between Unity and USD
+                var zFlippedVertex = new Vector3(vertex.x, vertex.y, -vertex.z);
```
vickycl, line 123: Much cleaner 🔥

vickycl (changes requested):
Much cleaner after refactoring, however the tests are failing. Will re-review once tests are fixed :)

package/com.unity.formats.usd/Dependencies/USD.NET.Unity/Geometry/PointInstancerSample.cs
lee-aandrew: Thanks! but its actually part of Jeremy's [fix](https://jira.unity3d.com/browse/USDU-305) 😆

package/com.unity.formats.usd/Tests/Common/BaseFixture.cs
```diff
@@ -30,7 +30,8 @@ public abstract class BaseFixture
         protected string ArtifactsDirectoryFullPath => Path.Combine(Application.dataPath, ArtifactsDirectoryName);
         protected string ArtifactsDirectoryRelativePath => Path.Combine("Assets", ArtifactsDirectoryName);
         protected string TestAssetDirectoryName => TestAssetData.Directory.FolderName;
-        protected string TestUsdAssetDirectoryRelativePath => Path.Combine("Packages", "com.unity.formats.usd", "Tests", "Common", "Data", TestAssetDirectoryName);
+        protected string TestPackageName => Directory.Exists(Path.Combine("Packages", "com.unity.formats.usd", "Tests")) ? "com.unity.formats.usd" : "com.unity.formats.usd.tests";
+        protected string TestUsdAssetDirectoryRelativePath => Path.Combine("Packages", TestPackageName, "Tests", "Common", "Data", TestAssetDirectoryName);
```
vickycl, line 34: Is this potentially why other tests use the GUID for the test assets rather than the path? I don't mind this approach, but if that's the way we work around it in other tests it would be good to be consistent here.

lee-aandrew: actually i think its just a lack of a proper convention / rule for this since
using GUID's still need access to "UnityEditor.AssetDatabase" which is unavailable for Runtime Standalone tests

I do want to go through all existing test cases and refactor them to be more consistent, but that will have to be done in a different PR

vickycl: Great, yeah it would be nice to remove the hardcoded paths. This is ok for now though :)

vickycl (approved):
Great job! Thanks for the iterations on the tests.

lee-aandrew:
Thanks for the reviews! @vickycl

PR #367, opened 2023-04-04
[Draft][USDU-331] Make package samples more user friendly

lee-aandrew:
## Purpose of this PR

**Ticket/Jira #:**

Currently our package samples can be a bit confusing for users,
To increase user friendly-ness the following have been addressed:
- In general, all Samples are much less Play Mode reliant - I've changed all samples to be done via Editor buttons, with changes being made in Edit Mode
- All Samples should be able to be fully executed within the "SAMPLE ENTRY POINT" object, without having to select different Unity GameObjects

- ExportAnimationWithRecorder (New):
-- Quick example scene with serialized fields that show the proper set up of how to record animation of Unity GameObject and export it as a USD file

- ExportMesh:
-- Made the process of exporting a more step-by-step process

- HelloUsd: 
-- Added more console debug logs
-- Added auto focus to console window

- ExportMeshTransformOverrides
-- Set up a scene where a USD asset is imported
-- Randomly changes the transform value
-- Exports it as a transform overrides USD file

- ExportMeshWithAnimation
-- Made it more a step by step process

- ImportMaterials
-- Made it more a step by step process

- ImportMesh
-- Made it more a step by step process

- ImportProcessor:
-- Added a new scene that show off the functionalities of CombineMeshes.cs and SetHideFlags.cs that implements Unity USD's ImportProcessor

- UsdTimelinePlayable:
-- Quick example scene with serialized fields that show the proper set up of how to play the animation of an imported USD asset

## Testing

## Overall Product Risks

**Complexity:**

**Halo Effect:**

## Additional information

**Note to reviewers:**

**Reminder:**


package/com.unity.formats.usd/Samples/UsdTimelinePlayable/README.md
```diff
@@ -0,0 +1,21 @@
+// Copyright 2017 Google Inc. All rights reserved.
```
CraigHutchinson: New file has the 'Google' copyright

package/com.unity.formats.usd/Samples/HelloUsd/HelloUsdExample.cs
```diff
@@ -29,8 +30,14 @@ class MyCustomData : SampleBase
 
         void Start()
         {
+#if UNITY_EDITOR
+            FocusConsoleWindow();
```
CraigHutchinson: `FocusConsoleWindow()` is a great addition, would be worth checking when the Console is not open. Presently appears to throw an exception and the sample may not continue?

lee-aandrew: According to [this doc](https://docs.unity3d.com/ScriptReference/Debug-developerConsoleVisible.html) if the Console window is not open, any Debug.Log call will open it so it should be fine, but I did still implement a try catch just in case something goes catastrophically wrong

package/com.unity.formats.usd/package.json
vickycl, line 1: We discussed this change in the past and decided not to do this, my bad, I was intending to document it but never got around to it. Described here: https://github.com/Unity-Technologies/usd-unity-sdk/issues/319#issuecomment-1351418030 and some discussion here: https://unity.slack.com/archives/C03R5ARENA1/p1670600438484989

Basically, this conversion is done automatically by the UPM-CI packing step, and shouldn't be committed like this. We decided to leave the Samples without the ~ for easier editing by package developers, because the ~ stops the folder from being imported and thus the sample are 'hidden' and have to be copied into the Assets folder using the import button in PackMan. With the current setup (no ~) I believe this means if an API is changed, we should get a compile error from the Samples and can update them, rather than the breakage being committed by accident, but I haven't seen this happen so I'm not entirely sure! The UPM-CI step then fixes it up so that users get the full polished experience, while users using the repo are still able to make changes to the samples directly.

vickycl (changes requested):
I've run through the samples and I think there's a confusing mix of samples running in playmode and edit mode. Because we don't really support USD at runtime, I think we should make all of the samples consistently run in Editor only and maybe use more buttons to trigger behaviours (which I think would be more similar to what users will probably want to do).

Let's have a call to discuss next steps :)

package/com.unity.formats.usd/package.json
vickycl: For reference: https://docs.google.com/document/d/1rmxGh6Z9gtbQlGUKCsVBaR0RyHvzq_gsWoYs6sttzYA/edit

package/com.unity.formats.usd/Samples/ExportMeshWithAnimation/Explode.cs
```diff
@@ -18,33 +18,25 @@ namespace USD.NET.Examples
...
-        }
+        private float currTime;
 
         void Update()
         {
-            if (!m_active)
-            {
-                return;
-            }
+            currTime += Time.deltaTime;
```
vickycl, line 30: What is the intention behind this change? This change will cause the explosion force to be added repeatedly, which seems odd to me.

package/com.unity.formats.usd/Samples/ExportMeshWithAnimation/ExportMeshWithAnimationExample.cs
```diff
@@ -62,10 +64,21 @@ public enum FrameRate
...
+        [HeaderAttribute("Export Related Variables")]
         // The root GameObject to export to USD.
         public GameObject m_exportRoot;
         public GameObject[] m_trackedRoots;
 
+        public bool m_exportMaterials = true;
+        public BasisTransformation m_convertHandedness = BasisTransformation.SlowAndSafe;
+        public ActiveExportPolicy m_activePolicy = ActiveExportPolicy.ExportAsVisibility;
+
+        // The path to where the USD file will be written.
```
vickycl: Worth specifying this is a file name, not a path, or alternatively keep it as an absolute path. Currently it's silently appended to the dataPath so it's not clear to the user that it would be placed in the Assets folder.

package/com.unity.formats.usd/Samples/ExportMeshWithAnimation/ExportMeshWithAnimationExample.cs
```diff
@@ -260,9 +266,11 @@ void LateUpdate()
             m_context.exportMaterials = false;
 
             // Exit once we've recorded all frames.
-            if (m_curFrame > m_frameCount)
+            if (m_curFrame > m_maxFrameCount)
             {
                 StopRecording();
+                IsFinishedRecording = true;
+                AssetDatabase.Refresh();
```
vickycl, line 229: Seems odd to do this during PlayMode. What's the reason for it?

package/com.unity.formats.usd/Samples/HelloUsd/Editor/HelloUsdExampleEditor.cs
```diff
@@ -0,0 +1,34 @@
...
+namespace Unity.Formats.USD.Examples
+{
+    [CustomEditor(typeof(HelloUsdExample))]
+    public class HelloUsdExampleEditor : Editor
+    {
+        public override void OnInspectorGUI()
+        {
+            DrawDefaultInspector();
+
+            HelloUsdExample combineMeshesScript = (HelloUsdExample)target;
```
vickycl: How come this is called 'Combine Meshes'?

package/com.unity.formats.usd/Samples/ImportProcessor/Editor/SetHideFlagsEditor.cs
```diff
@@ -0,0 +1,32 @@
...
+    public class SetHideFlagsEditor : Editor
+    {
+        public override void OnInspectorGUI()
+        {
+            DrawDefaultInspector();
+
+            SetHideFlags combineMeshesScript = (SetHideFlags)target;
+            if (GUILayout.Button("Click Me to Refresh USD Asset"))
+            {
+                combineMeshesScript.RefreshUSD();
```
vickycl: Ahh, looks like a remnant of copy-paste. Can you change these to more accurate variable names please?

package/com.unity.formats.usd/Samples/UsdTimelinePlayable/README.md
vickycl: I think we probably shouldn't add new files with that Google copyright, is that what you're getting at Craig?

vickycl (changes requested):
A few comments based only on the PR, let me know when you want me to check out and test the samples :)

package/com.unity.formats.usd/Samples/ExportMeshWithAnimation/Explode.cs
lee-aandrew: yes its to do the explode animation repeatedly - i made the change so that if the User wants to record the explode animation but misses the explode animation they dont have to restart the scene, and repeat the export steps from the beginning

package/com.unity.formats.usd/Samples/ExportMeshWithAnimation/ExportMeshWithAnimationExample.cs
lee-aandrew: I wanted to refresh the Asset Screen right away as soon as the recording part was done in case the user is expecting it as soon as its done recording - but we could move it to do this when the scene playmode is stopped, what do you think?

package/com.unity.formats.usd/Samples/HelloUsd/Editor/HelloUsdExampleEditor.cs
lee-aandrew: copy pasta error 😅

package/com.unity.formats.usd/Samples/ExportMesh/Editor/ExportMeshExampleEditor.cs
```diff
@@ -25,30 +25,80 @@ public override void OnInspectorGUI()
...
+                        }
+
+                        if (GUILayout.Button("3. Set up Export Context"))
+                        {
+                            SampleUtils.FocusConsoleWindow();
+
+                            script.SetUpExportContext();
+                        }
+
+                        if (GUILayout.Button("3. Set up Export Context"))
```
vickycl, line 53: Remove duplicated button 3

vickycl: Only some of the buttons generated a log, which made it seem like the buttons that didn't were missing something. I think for consistency, within the same sample all of the buttons should do the same behaviour :)

package/com.unity.formats.usd/Samples/ExportMesh/Editor/ExportMeshExampleEditor.cs
```diff
@@ -25,30 +25,80 @@ public override void OnInspectorGUI()
...
+                case ExportMeshExample.UsdFileExtension.usd:
+                case ExportMeshExample.UsdFileExtension.usda:
+                case ExportMeshExample.UsdFileExtension.usdc:
+                    {
+                        if (GUILayout.Button("2. Create New USD Scene"))
+                        {
+                            SampleUtils.FocusConsoleWindow();
+
+                            script.CreateNewUsdScene();
+                            Debug.Log($"<color={SampleUtils.TextColor.Green}>Created USD file: <b><{script.m_newUsdFileName}.{script.FileExtension}></b> under project <b>'{SampleUtils.SampleArtifactRelativeDirectory}'</b> folder</color>");
```
vickycl: I like the colored text to indicate it's come from the sample. Can we do this everywhere? for example, the logs from ExportMeshTransformOverrides aren't the same.

Instead of rewriting` <color={SampleUtils.TextColor.Green}> [text] </color>` you could make this a utility function that does that for you.

package/com.unity.formats.usd/Samples/ImportProcessor/PostProcessComponents_CombineMeshes/ImportProcessorExample_PostProcessComponents.cs
vickycl, line 1: I found the layout of this sample confusing when I tried it out. Can we rename the scripts that are doing the actual PostProcess work back to something that describes what they're doing (such as the previous 'CombineMeshes' and 'SetHideFlags'), and have both the samples in the same scene? We could have a reset button to facilitate running one and then the other. Also the USD asset is nested under the entry point in the hierarchy instead of under 'World' like it is on others, which was harder to find. 

Having the common editor script also means it isn't as explanatory what the buttons are doing. 'Refresh Asset' could be more descriptive of how it's reimporting the asset to trigger whichever post processing is running, for example.

package/com.unity.formats.usd/Samples/ImportProcessor/ExampleAsset.meta
vickycl, line 1: I think so that it's easier to see, let's bring the MultipleMeshes.usd file into the top level ImportProcessor folder for the sample instead of this extra folder level.

package/com.unity.formats.usd/Samples/ImportMaterials/Editor/ImportMaterialsExampleEditor.cs
```diff
@@ -0,0 +1,68 @@
...
+    [CustomEditor(typeof(ImportMaterialsExample))]
+    public class ImportMaterialsExampleEditor : Editor
+    {
+        public override void OnInspectorGUI()
+        {
+            DrawDefaultInspector();
+
+            ImportMaterialsExample script = (ImportMaterialsExample)target;
+
+            if (GUILayout.Button("1. Initialize USD Package"))
```
vickycl: This should be 'USD plugin' or just 'USD' I think, not package

package/com.unity.formats.usd/Samples/ImportMaterials/Editor/ImportMaterialsExampleEditor.cs
```diff
@@ -0,0 +1,68 @@
...
+            {
+                script.InitializeUsd();
+            }
+
+            if (GUILayout.Button("2. Initialize Sample Shader Map Dictionary"))
+            {
+                script.InitializeSampleShaderMapDictionary();
```
vickycl, line 38: I don't think it makes sense to do each of these steps separately. They aren't descriptive of what they're doing or why they're needed, and if a user does them out of order it results in errors. I think it would be reasonable to put them all into one step of 'Import USD with Materials', and a second step to enter playmode to show the material (like the sample originally did). There could be options on the Editor script to change material import mode, but with Preview Material set by default.

package/com.unity.formats.usd/Samples/ImportMaterials/ImportMaterialsExample.cs
```diff
@@ -44,31 +47,14 @@ public struct ShaderPair
...
-        {
-            if (!m_cube)
-            {
-                return;
-            }
-
-            var rot = m_cube.localEulerAngles;
-            rot.y += Time.deltaTime * 10 + Mathf.Sin(Time.time) * .1f + .05f;
-            m_cube.localEulerAngles = rot;
-        }
```
vickycl, line 59: I quite liked the cube that showed off the imported material, what was the reasoning for removing it?

vickycl (changes requested):
ImportMesh samples are much nicer now :) There's just some small points on those ones for consistency.

I found the new layout for the PostProcessing samples to be confusing, as a couple of things were hidden away in folder structures or different scenes. Hopefully the comments made sense?

For the ImportMaterials samples, I think the step breakdown is too much. There's no value to having each step as a separate button, and more chance of breakage. I think we should return that to being more simple, as a user wanting to emulate that sample would need to look at the code anyway and not care about the button steps. I also think it's missing the previous behaviour of applying the imported material to the cube in the scene and rotating in playmode to show it off. 

There are two missing .sample.json files- I think they were for ExportMesh and ExportMeshTransformOverrides. These files help the UPM-CI find the Samples and add them to the package json.

package/com.unity.formats.usd/Samples/UsdTimelinePlayable/Editor/UsdTimelineRecorderExampleEditor.cs
```diff
@@ -0,0 +1,59 @@
...
+            }
+#else
+            var script = (UsdTimelineRecorderExample)target;
+
+            if (!script.m_playableDirectorExists)
+            {
+                GUILayout.Label($"First you need to:");
+                if (GUILayout.Button("Add 'Playable Director' Component"))
+                {
+                    script.AddPlayableDirectorComponent();
```
vickycl: It's simpler to do this all through the Timeline window, as you get the matching asset created. I think it makes more sense to direct users to that.

package/com.unity.formats.usd/Samples/UsdTimelinePlayable/Editor/UsdTimelineRecorderExampleEditor.cs
```diff
@@ -0,0 +1,59 @@
...
+                GUILayout.Label($"  4. Within the 'Timeline Window' Right Click within any <b>empty area</b> of the window, and select Unity.Formats.USD > <color={SampleUtils.TextColor.Yellow}>'Usd Playable Track'</color>.", labelStyle);
+                GUILayout.Label($"  5. Set the <color={SampleUtils.TextColor.Blue}>Source Object</color> within the newly created <b>'Usd Playable Track'</b> as the USD GameObject, <color={SampleUtils.TextColor.Yellow}>'UsdAssetWithTimeline'</color> (first child of this object).", labelStyle);
+                GUILayout.Label($"  6. Within the 'Timeline Window' Right Click within the newly created <b>Usd Playable Track</b>, and select <color={SampleUtils.TextColor.Yellow}>'Add From Usd Asset'</color>.", labelStyle);
+                GUILayout.Label($"  7. In the new pop up <color={SampleUtils.TextColor.Blue}>'Select Usd Asset'</color> window select the USD GameObject, <color={SampleUtils.TextColor.Yellow}>'UsdAssetWithTimeline'</color>.", labelStyle);
+                GUILayout.Label($"  8. You can now play the Timeline animation and see the animation being played in your <b>Scene View</b> window.", labelStyle);
```
vickycl: This is only meant to be a 'Sample' so that a user can play around with the features, not a full walkthrough. Having the steps written out makes it quite brittle and easy to become out of date. I personally think these steps would be better suited to being in the documentation instead of a Sample, where they will be easier and more obvious to keep up to date.

package/com.unity.formats.usd/Samples/UsdTimelinePlayable/UsdTimelineRecorderExample.cs
```diff
@@ -0,0 +1,24 @@
...
+{
+    [ExecuteAlways]
+    public class UsdTimelineRecorderExample : MonoBehaviour
+    {
+        [HideInInspector]
+        public bool m_playableDirectorExists = false;
+
+        private void Update()
+        {
+            m_playableDirectorExists = this.GetComponent<PlayableDirector>() != null;
```
vickycl: I've mentioned above that I don't think we should be doing this as part of a sample, but just a bit of extra advice here- We shouldn't do GetComponent in an Update loop, even if it is 'just' in the Editor. GetComponent can be very costly and as the Update loop runs every frame this can be harmful for performance, so we don't want to encourage it to users :)

vickycl (changes requested):

package/com.unity.formats.usd/Samples/ImportMaterials/ImportMaterialsExample.cs
lee-aandrew: Yeah i guess i could add it back in to show it off

package/com.unity.formats.usd/Samples/UsdTimelinePlayable/README.md
lee-aandrew: I think the Google Copyright is from when Jeremy was at Google - I believe he started this project when he was part of Google and i suspect he copy-pasta'd some of his scripts

But having said that, I'll be sure to set all of these to "2023 Unity Technologies"

package/com.unity.formats.usd/Samples/UsdTimelinePlayable/Editor/UsdTimelineRecorderExampleEditor.cs
lee-aandrew: I've removed a lot of steps from both this and ExportAnimationWithRecorderExample with the belief that all the serialized fields will be saved 🤞 

Lemme know what you think!

package/com.unity.formats.usd/Samples/ExportMesh/Editor/ExportMeshExampleEditor.cs
```diff
@@ -25,30 +25,77 @@ public override void OnInspectorGUI()
...
+                            script.CreateNewUsdScene();
+                            Debug.Log(SampleUtils.SetTextColor(SampleUtils.TextColor.Green, $"Created USD file: <b><{script.m_newUsdFileName}.{script.FileExtension}></b> under project <b>'{SampleUtils.SampleArtifactRelativeDirectory}'</b> folder"));
+                        }
+
+                        if (GUILayout.Button("3. Set up Export Context"))
+                        {
+                            SampleUtils.FocusConsoleWindow();
+
+                            script.SetUpExportContext();
+                            Debug.Log($"Export Context has been set up.");
```
vickycl, line 58: I think each of these debug.logs should use the same green, so that they all stand out as behaviour caused by the samples

package/com.unity.formats.usd/Samples/ExportMesh/Editor/ExportMeshExampleEditor.cs
```diff
@@ -25,30 +25,77 @@ public override void OnInspectorGUI()
...
+
+                            script.SaveScene();
+                            AssetDatabase.Refresh();
+                            Debug.Log(SampleUtils.SetTextColor(SampleUtils.TextColor.Green, $"Data exported to <b><{script.m_newUsdFileName}.{script.FileExtension}></b> has been saved."));
+                            Debug.Log($"The file <b><{script.m_newUsdFileName}.{script.FileExtension}></b> is available at your project <b>'{SampleUtils.SampleArtifactRelativeDirectory}'</b> directory.");
+                        }
+
+                        if (GUILayout.Button("6. Close Scene"))
+                        {
+                            script.CloseScene();
```
vickycl, line 81: I don't see a reason to do the Save and Close steps seperately- I think they should be combined into one step to reduce clicks.

package/com.unity.formats.usd/Samples/ExportMesh/Editor/ExportMeshExampleEditor.cs
vickycl: 'Setup Export Context' isn't really a useful step. It doesn't explain what is happening or why. Perhaps just combine this with the export step, or at least call it "Generate settings for export"

package/com.unity.formats.usd/Samples/ExportMeshWithAnimation/Editor/ExportMeshWithAnimationExampleEditor.cs
```diff
@@ -0,0 +1,70 @@
...
+                if (GUILayout.Button("Enter playmode and start animation"))
+                {
+                    EditorApplication.isPlaying = true;
+                }
+            }
+            else
+            {
+                if (script.IsFinishedRecording)
+                {
+                    if (GUILayout.Button("Recording Complete - Stop Scene"))
```
vickycl, line 40: Should be "Recording Complete - Exit Playmode"

package/com.unity.formats.usd/Samples/ExportMesh/.sample.json
```diff
@@ -1,4 +1,4 @@
 {
     "displayName" : "ExportMesh",
-    "description" : "A sample showing how to record a scene to a USD file."
+    "description" : "A sample showing how Unity Game Object can be exported as a USD file (without animation)."
```
vickycl, line 3: Add something along the lines of "using the Unity USD APIs directly" so that it's clear that this isn't the only way of doing it for a first time user

package/com.unity.formats.usd/Samples/ExportMeshTransformOverrides/.sample.json
```diff
@@ -0,0 +1,4 @@
+{
+    "displayName" : "ExportMeshTransformOverrides",
+    "description" : "A sample showing an example of imported USD file exported as transform override."
```
vickycl, line 3: I think this should be "A sample showing how to export an imported USD asset with modified transforms as a USD override file using the Unity USD APIs."

package/com.unity.formats.usd/Samples/ExportMeshWithAnimation/.sample.json
```diff
@@ -0,0 +1,4 @@
+{
+    "displayName" : "ExportMeshWithAnimation",
+    "description" : "A sample showing how to record a scene to a USD file."
```
vickycl, line 3: Should be "A sample showing how to record an animation to a USD file without using the recorder package."

package/com.unity.formats.usd/Samples/ImportMesh/Editor/ImportMeshExampleEditor.cs
```diff
@@ -0,0 +1,67 @@
...
+        public override void OnInspectorGUI()
+        {
+            base.DrawDefaultInspector();
+            var script = (ImportMeshExample)target;
+            if (GUILayout.Button("1. Initialize USD"))
+            {
+                script.InitializeUsd();
+            }
+
+            if (GUILayout.Button("2. Set USD File Path..."))
```
vickycl, line 33: This additional step makes the sample confusing. To make it less confusing and also remove the risk of it going wrong, just remove it and always import the USD file provided with the Sample.

package/com.unity.formats.usd/Samples/ImportProcessor/Editor/ImportProcessorExample_EntryPointEditor.cs
vickycl, line 1: I think this approach is too confusing and detracts away from the automatic post processing we're trying to demonstrate. I think in the interest of getting this landed we should revert this sample to how it was for now, but I've sent you a message on Slack so we can discuss.

package/com.unity.formats.usd/Samples/UsdTimelinePlayable/SetDefaultUSDMesh.cs
```diff
@@ -21,18 +21,25 @@ namespace Unity.Formats.USD.Examples
...
-            if (string.IsNullOrEmpty(asset.usdFullPath))
-            {
-                asset.usdFullPath = Path.Combine(PackageUtils.GetCallerRelativeToProjectFolderPath(), K_DEFAULT_MESH);
-            }
+            SetSampleUsdFilePath();
+        }
+
+        void Update()
+        {
+            SetSampleUsdFilePath();
```
vickycl, line 36: This shouldn't be called every frame :)

vickycl (changes requested):

package/com.unity.formats.usd/Samples/UsdTimelinePlayable/SetDefaultUSDMesh.cs
lee-aandrew: I think having this run every frame is beneficial to make sure the user cant change the path for this USD Asset to another USD file since not every USD file will contain animation
I just want to enforce this specific asset on this Sample

package/com.unity.formats.usd/Samples/HelloUsd/Unity.Formats.USD.HelloUsd.asmdef
vickycl, line 1: [hijacking this file to say] I think the Hello USD would benefit from calling out in the .sample.json that it is the introductory walkthrough of how to write and read a USD file through code, so that it's easy for a user to pick a first sample to look at and also understand why it's broken down to be so granular.

package/com.unity.formats.usd/Samples/UsdTimelinePlayable/SetDefaultUSDMesh.cs
vickycl: The file path isn't easily modifiable anyway, and it's demonstrating a bad practice for a user (running something every editor frame, especially something doing a GetComponent and a string manipulation). It would be a very edge case behaviour if a user modified that file path manually, and personally if they did so I think it would be better if it caused an issue (as it would in a real case) instead of silently 'fixed' itself

PR #370, opened 2023-04-14
[USDU-374] Export omits EditorOnly tag objects

lee-aandrew:
## Purpose of this PR
Unity has a Tag called "EditorOnly", if objects in Unity is set with this tag, it should only be visible / interactable within the Unity Editor

This means USD Export should not include any objects with the tag "EditorOnly"

Issue reported [here](https://github.com/Unity-Technologies/usd-unity-sdk/issues/303)

**Ticket/Jira #:**
[USDU-374](https://jira.unity3d.com/browse/USDU-374)

## Testing

PR for test case(s):
https://github.com/Unity-Technologies/usd-unity-sdk/pull/342

## Overall Product Risks

**Complexity:**

**Halo Effect:**

## Additional information

**Note to reviewers:**

**Reminder:**

### DIFF
--- a/package/com.unity.formats.usd/Runtime/Scripts/IO/Scene/SceneExporter.cs
+++ b/package/com.unity.formats.usd/Runtime/Scripts/IO/Scene/SceneExporter.cs
@@ -245,6 +245,11 @@ namespace Unity.Formats.USD
                     continue;
                 }
 
+                if (go.CompareTag("EditorOnly"))
+                {
+                    continue;
+                }
+
                 if (go != root && !go.transform.IsChildOf(root.transform))
                 {
                     continue;
--- a/package/com.unity.formats.usd/Tests/Editor/USDPayloadComponentTests.cs
+++ b/package/com.unity.formats.usd/Tests/Editor/USDPayloadComponentTests.cs
@@ -207,6 +207,7 @@ namespace Unity.Formats.USD.Tests
             Assume.That(usdPayload.IsLoaded, Is.False, "UsdPayload.IsLoaded should be set to false.");
         }
 
+        [Ignore("[USDU-329] Unstable test result")]
         [UnityTest]
         public IEnumerator ChangingUsdPayloadStateFromUnloadedToLoaded_MarksSceneDirty()
         {
@@ -224,6 +225,7 @@ namespace Unity.Formats.USD.Tests
             Assert.IsTrue(m_UnityScene.isDirty, "Scene should be dirty after changing UsdPayload objects");
         }
 
+        [Ignore("[USDU-329] Unstable test result")]
         [UnityTest]
         public IEnumerator ChangingUsdPayloadStateFromLoadedToUnloaded_MarksSceneDirty()
         {


vickycl (changes requested):
On reflection, I don't think this is the right approach. The intention of the Editor Only flag is to strip authoring objects at Build time of a player. However, as USD is currently considered an authoring tool in our package, I would expect objects intended for in-editor authoring to persist in a roundtrip, for example. I think this is something we should review when we work on improving export, as it should definitely be an option for users, but I would prefer the default to be too many objects being exported, rather than some objects being secretly removed. What are your thoughts on this?

lee-aandrew:
> On reflection, I don't think this is the right approach. The intention of the Editor Only flag is to strip authoring objects at Build time of a player. However, as USD is currently considered an authoring tool in our package, I would expect objects intended for in-editor authoring to persist in a roundtrip, for example. I think this is something we should review when we work on improving export, as it should definitely be an option for users, but I would prefer the default to be too many objects being exported, rather than some objects being secretly removed. What are your thoughts on this?

I think one can also argue that "EditorOnly" means Unity Editor only, and there maybe objects you add while working on Unity that you dont necessarily want to see in DCCs or USDViewer on export - and considering how this was an issue from a user, i guess there is an expectation of this behaviour for this

I think the best option would have been to keep data about Unity tags / layer (maybe via custom value similar to what HelloUSDExample.cs is showing off - if tag is set to "EditorOnly" we save it as such and set visible/enabled to false) so that if an object is exported through Unity, and then imported, the data is kept - at least for the default options we have for tags / layers

vickycl:
> I think one can also argue that "EditorOnly" means Unity Editor only, and there maybe objects you add while working on Unity that you dont necessarily want to see in DCCs or USDViewer on export - and considering how this was an issue from a user, i guess there is an expectation of this behaviour for this

I agree that in this case 'Editor only' is vague, but it was done at a time where exporting GameObjects back to other DCCs wasn't really a thing :) To users who view this tag the other way - as an 'authoring' tag - this will be a regression, where 'authoring only' objects they expect to see in other DCCs no longer appear, and do not come back into the Editor on round tripping. In my view it is better for these objects to still be exported to USD where they can be ignored or removed, than to secretly be removed and not appear at all. I think we need a fully fleshed out solution that gives users the option, so that they can suit their workflow.

> I think the best option would have been to keep data about Unity tags / layer (maybe via custom value similar to what HelloUSDExample.cs is showing off - if tag is set to "EditorOnly" we save it as such and set visible/enabled to false) so that if an object is exported through Unity, and then imported, the data is kept - at least for the default options we have for tags / layers

This sounds like a good idea. Let's explore this as part of rewriting export :)

lee-aandrew:
> I agree that in this case 'Editor only' is vague, but it was done at a time where exporting GameObjects back to other DCCs wasn't really a thing :) To users who view this tag the other way - as an 'authoring' tag - this will be a regression, where 'authoring only' objects they expect to see in other DCCs no longer appear, and do not come back into the Editor on round tripping. In my view it is better for these objects to still be exported to USD where they can be ignored or removed, than to secretly be removed and not appear at all. I think we need a fully fleshed out solution that gives users the option, so that they can suit their workflow.

Makes sense, it would be weird if i export and re-import right away and see that the EditorOnly objects are now missing
I will close the PR!

PR #371, opened 2023-04-17
Rewrite of documentation

vickycl:
## Purpose of this PR

- Extracts import, export and 'USD in the Editor' into separate pages
- Adds missing details in several areas
- Changed some images and removed duplicates
- Removed a lot of detail from the Readme and point to the Documentation entry point instead, to reduce maintenance burden

## Testing

**Functional Testing status:** N/A

## Overall Product Risks

**Complexity:** 0

**Halo Effect:** 0

## Additional information

**Note to reviewers:**

**Reminder:**


package/com.unity.formats.usd/Documentation~/Images/usd-import-options.PNG
lee-aandrew, line 1: I just realized i have never used "Unload Subtree" and "Load Subtree" and it seems like it unloads / loads all payloads within an imported USD object

I feel like we should either
1. Explain what these buttons do (and maybe even rename them to be consistent with Load All / Unload All payloads?)
2. Or delete them from the menu since we already have a place for this function in the USD Asset component

package/com.unity.formats.usd/Documentation~/USD Export.md
```diff
@@ -0,0 +1,77 @@
...
+over "myCube"
+{
+    ...
+}
+```
+
+
+Note: Modifications to the transform of the Root GameObject are not currently reflected in the override, as Unity assumes the root in all USD files is at the origin.
+
+## Export Selected as USDZ
```
lee-aandrew, line 50: we should note that Unity GameObjects set with "EditorOnly" tag will be ignored for export

package/com.unity.formats.usd/Documentation~/USD Export.md
```diff
@@ -0,0 +1,77 @@
+# Exporting USD files
```
lee-aandrew, line 1: we should note that Unity GameObjects set with "EditorOnly" tag will be ignored for export

package/com.unity.formats.usd/Documentation~/USD.md
```diff
@@ -1,126 +1,114 @@
...
 ![USD unlimited_weights](Images/USD_unlimited_weights.png)
 
 ## Requirements
 
 * Windows / OSX (Intel only)
 * Unity version: 2019.4 and up
-* API Compatibility Level .NET 4.x is no longer required but still provides better performances than .NET 2   
-   In Edit > Project Settings > Player :   
-    ![USD .NET version](Images/USD_.NET_version.png)
+* API Compatibility Level .NET 4.x is no longer required but still provides better performances than .NET 2
```
CraigHutchinson: This doesn't read as a requirement - Could simplify the wording as  'Recommendation' or add a section for that.

package/com.unity.formats.usd/Documentation~/USD Export.md
```diff
@@ -0,0 +1,77 @@
...
+
+* USDC (.usd or .usdc)
+* USDA (.usda)
+* USDZ (.usdz)
+
+The default file format when exporting is .usd with USDC encoding, but changing the file ending in the save menu to .usda will export as USDA.
+
+## Export Selected with Children
+
+On Mac you will see multiple file format options- USDC and USDA.
```
lee-aandrew: we could probably add a space between the "options" and "-"
though now im curious if what you did is the correct form in the UK

package/com.unity.formats.usd/Documentation~/USD.md
```diff
@@ -1,126 +1,114 @@
...
+
+* or browsing for a local package while working with source.
+
+Once the USD package is installed, a USD menu will appear, enabling you to easily import and export USD files.
 
 ![USD menu](Images/USD_menu.png)
 
-The USD importer works with linear color space only. To ensure colors are imported correctly,
-set the project color space to "linear" in Edit > Project Settings > Player:
+The USD importer works with linear color space only. To ensure colors are imported correctly, set the project color space to "linear" in Edit > Project Settings > Player:
```
CraigHutchinson: Sounds like a Requirement, could be added to the relevant section.

package/com.unity.formats.usd/Documentation~/USD.md
```diff
@@ -1,126 +1,114 @@
...
-
-### Exporting via Usd Recorder Track (Legacy)
-
-When compatibility with runtime is required (i.e for a standalone build), the recommended option is to use the USD package Recorder Track :
-* From the Timeline window, right-click and select **Unity.Formats.USD > Usd Recorder Track**,
-* Then right-click on the track and **add USD Recorder Clip**.
-
->  **Note:** This feature has no dependency to and is not based on the Recorder package.
-
+* Importing and exporting USD:
```
CraigHutchinson, line 63: Maintenance - the list of features is duplicated in two places. (Here &  top-level readme.) Makes sense to have this for the repo & package cases though.

 Without a preprocessor tool (Mustache or similar) we need to make sure to manually replicate both of these so good job doing so here. For interest, splitting to a separate  features 'include' file is also not viable as Security concerns has always blocked this feature (see https://github.com/github/markup/issues/172#issuecomment-33241601)

package/com.unity.formats.usd/Documentation~/USD.md
```diff
@@ -1,126 +1,114 @@
...
 
 ## Requirements
 
 * Windows / OSX (Intel only)
 * Unity version: 2019.4 and up
-* API Compatibility Level .NET 4.x is no longer required but still provides better performances than .NET 2   
-   In Edit > Project Settings > Player :   
-    ![USD .NET version](Images/USD_.NET_version.png)
+* API Compatibility Level .NET 4.x is no longer required but still provides better performances than .NET 2 
+In Edit > Project Settings > Player : ![USD .NET version](Images/USD_.NET_version.png)
```
lee-aandrew: Seems like you deleted 
Images/USD_.NET_version.png

in this PR

package/com.unity.formats.usd/Documentation~/USD.md
```diff
@@ -1,126 +1,114 @@
...
+        * Formats: USDA, USDC, USDZ
+        * Export of Transform Overrides
+    * Export via Recorder package
+* Composition:
+    * Variant Selection
+    * Payloads
+        * Load All and Individual
+    * Layer stacks
+* Geometry:
+    * UV Set
```
lee-aandrew, line 76: I kind of want to specify somewhere in the README that viewing UV channels within Unity is available on Unity 2020+
Just like as a reminder

What do you think?

package/com.unity.formats.usd/Documentation~/USD.md
```diff
@@ -1,126 +1,114 @@
...
+* Geometry:
+    * Multiple UV Sets
+    * Importing mesh UVs for a prim without a material attached
+* Composition:
+    * Purposes
+* Primitive Types:
+    * Camera
+        * We do not currently import and export all physical camera settings
+    * Materials:
+        * Transparency Settings (these can be set manually)
```
lee-aandrew: Transparency & Double-Sided Settings

package/com.unity.formats.usd/Documentation~/USD.md
```diff
@@ -1,126 +1,114 @@
...
+
+Once the USD package is installed, a USD menu will appear, enabling you to easily import and export USD files.
 
 ![USD menu](Images/USD_menu.png)
 
-The USD importer works with linear color space only. To ensure colors are imported correctly,
-set the project color space to "linear" in Edit > Project Settings > Player:
+The USD importer works with linear color space only. To ensure colors are imported correctly, set the project color space to "linear" in Edit > Project Settings > Player:
 
 ![USD linear](Images/USD_linear.png)
```
lee-aandrew: "Images/USD_linear.png" has been deleted in this PR

package/com.unity.formats.usd/Documentation~/USD.md
```diff
@@ -1,126 +1,114 @@
...
+
+* installing by name, "com.unity.formats.usd";
+
+* installing by Git URL, "https://github.com/Unity-Technologies/usd-unity-sdk.git?path=package/com.unity.formats.usd";
+
+* or browsing for a local package while working with source.
+
+Once the USD package is installed, a USD menu will appear, enabling you to easily import and export USD files.
 
 ![USD menu](Images/USD_menu.png)
```
lee-aandrew: "USD_menu.png" has been deleted in this PR

package/com.unity.formats.usd/Documentation~/USD.md
```diff
@@ -1,126 +1,114 @@
...
+The USD importer works with linear color space only. To ensure colors are imported correctly, set the project color space to "linear" in Edit > Project Settings > Player:
 
 ![USD linear](Images/USD_linear.png)
 
-In Unity 2019, the USD importer supports importing unlimited weights per vertex. However,
-to see the effect of more than 4 weights per vertex, this must be enabled in the project
-settings under Edit > Project Settings > Quality:
+In Unity 2019, the USD importer supports importing unlimited weights per vertex. However, to see the effect of more than 4 weights per vertex, this must be enabled in the project settings under Edit > Project Settings > Quality:
 
 ![USD unlimited_weights](Images/USD_unlimited_weights.png)
```
lee-aandrew: "Images/USD_unlimited_weights.png" has been deleted in this pr

package/com.unity.formats.usd/Documentation~/USD.md
```diff
@@ -1,126 +1,114 @@
...
-
-When compatibility with runtime is required (i.e for a standalone build), the recommended option is to use the USD package Recorder Track :
-* From the Timeline window, right-click and select **Unity.Formats.USD > Usd Recorder Track**,
-* Then right-click on the track and **add USD Recorder Clip**.
-
->  **Note:** This feature has no dependency to and is not based on the Recorder package.
-
+* Importing and exporting USD:
+    * Import as GameObject, Prefab, or Timeline Clip
+        * Formats: USDA, USDC, USDZ
```
lee-aandrew: "USD" format is omitted here 😄

package/com.unity.formats.usd/Documentation~/USD.md
```diff
@@ -1,126 +1,114 @@
...
+    * Export Game Objects to USD
+        * Formats: USDA, USDC, USDZ
```
lee-aandrew: "USD" format is omitted here as well

package/com.unity.formats.usd/Documentation~/USD Import.md
```diff
@@ -0,0 +1,53 @@
...
+To import a USD file into Unity, use the USD context menu, and select one of the 'Import as ...' options as required. Supported USD file formats are:
+
+-   .USD
+-   .USDA
+-   .USDC
+-   .USDZ
+
+## Importing Options
+
+-   **Import As GameObjects**: Imports selected USD file as GameObjects directly into the Scene Hierarchy.
```
CraigHutchinson: Would be good to elaborate on the benefits of the GameObjects vs Prefab option? As a 'new user' I'm really not clear which option should be my go-to option to start with. Will I find a limitation from one vs the other etc.

package/com.unity.formats.usd/Documentation~/USD Import.md
```diff
@@ -0,0 +1,53 @@
...
+### Meshes
+
+Convert basis:
+- As above, meshes also go through a basis conversion as required.
+
+Triangulate meshes:
+- Converting from USD-supported mixed face meshes to a triangular mesh.
+
+Change vertex winding order: 
+- Unity uses a clockwise mesh winding order, where other DCCs may use anti-clockwise.
```
CraigHutchinson: Localization: American English would prefer `counterclockwise` (https://books.google.com/ngrams/graph?content=counterclockwise,anticlockwise&year_start=1900&year_end=2000&corpus=17&smoothing=3&share=) while British speakers are indifferent - I hear Unity docs prefer US standardization?

package/com.unity.formats.usd/Documentation~/USD Export.md
```diff
@@ -0,0 +1,77 @@
...
+
+On Windows the default is .usd with a USDC format, but you can change the file ending in the Save window to specify a different file ending and encoding.
+
+
+## Export Transform Overrides
+
+Modifications to transforms in USD files can be exported as overrides to the original files. 
+
+Overrides can be exported from the USD menu with the GameObject containing the [UsdAsset](/USD-in-the-Editor#USD-Asset) component selected, which will export transform overrides for the entire hierarchy. Alternatively, you can also export just the overrides when exporting from the Recorder window by changing the 'Override Setting' to 'Export Transform Overrides Only'.
```
CraigHutchinson: Dead-link: `/USD-in-the-Editor#USD-Asset` is does not work from Github (https://github.com/Unity-Technologies/usd-unity-sdk/blob/documentation/rewrite/package/com.unity.formats.usd/Documentation~/USD%20Export.md)

I think removing the leading '/' (root folder) shall resolve this as the file is in the same folder.

lee-aandrew (changes requested):
Added some comments - but awesome changes! 🏆

package/com.unity.formats.usd/Documentation~/USD Export.md
```diff
@@ -0,0 +1,77 @@
...
+## Export Selected with Children
+
+On Mac you will see multiple file format options- USDC and USDA.
+
+On Windows the default is .usd with a USDC format, but you can change the file ending in the Save window to specify a different file ending and encoding.
+
+
+## Export Transform Overrides
+
+Modifications to transforms in USD files can be exported as overrides to the original files.
```
CraigHutchinson, line 20: Sounds like a limitation for 'known limitations', I'm not sure? Is this maybe this is just the more 'general' limitation that Unity and USD have been developed with differing goals I guess.

package/com.unity.formats.usd/Documentation~/USD.md
```diff
@@ -1,126 +1,114 @@
...
-to see the effect of more than 4 weights per vertex, this must be enabled in the project
-settings under Edit > Project Settings > Quality:
+In Unity 2019, the USD importer supports importing unlimited weights per vertex. However, to see the effect of more than 4 weights per vertex, this must be enabled in the project settings under Edit > Project Settings > Quality:
 
 ![USD unlimited_weights](Images/USD_unlimited_weights.png)
 
 ## Requirements
 
 * Windows / OSX (Intel only)
 * Unity version: 2019.4 and up
```
CraigHutchinson: Sanity-check: Is there no upper-limit on supported Unity version?

CraigHutchinson (changes requested):
One necessary change:
- There is an internal broken link to `/USD-in-the-Editor#USD-Asset`

Other than that looks good, made observations which may be applied in future work as desired 👍

package/com.unity.formats.usd/Documentation~/USD.md
lee-aandrew: I haven't done a full testing on our trunk, but yes the package works on latest unity

package/com.unity.formats.usd/Documentation~/USD.md
vickycl: That's not a specification of this package but of USD, so I don't think this is the right place :)

package/com.unity.formats.usd/Documentation~/USD.md
vickycl: All of the images in the `Images` folder were duplicated in `package/com.unity.formats.usd/Documentation~/Images`, so I've only removed the duplicates at the higher level. This path is relative so *should* use the remaining one.

package/com.unity.formats.usd/Documentation~/USD.md
vickycl: As above, this points to the one in the documentation folder.

package/com.unity.formats.usd/Documentation~/USD.md
vickycl: I was going for the encoding format rather than the file endings here... a .usd file just defaults to USDC encoding I believe

package/com.unity.formats.usd/Documentation~/Images/usd-import-options.PNG
vickycl: The only benefit I can guess at of this over the UsdPayload tickbox is that you don't have to manually find all of your payloads to load them as long as they are below a common ancestor. Theoretically if you wanted to load all of the payloads under an 'environment' parent but not the ones under a 'character' parent without finding each individual payload (and that parent isn't itself a payload) then you could use this. You're right, it should be documented, and possibly have 'Payload' in the name for clarity.

package/com.unity.formats.usd/Documentation~/USD Export.md
vickycl: I'm not sure what you mean by this being a limitation? Exporting the Transform Overrides is a different import option to plain export, exporting only the USD 'Overs' for a USD file instead of the full file. It's a USD composition concept to allow layering. Does this answer your question?

package/com.unity.formats.usd/Documentation~/USD Export.md
vickycl: Thanks!

package/com.unity.formats.usd/Documentation~/USD Import.md
vickycl: Ack. Didn't even clock that (geddit?) Thanks for pointing it out!

package/com.unity.formats.usd/Documentation~/USD.md
vickycl: Just in case there's other things we're missing that we don't even know about, I'll keep this vague and put them as examples?

lee-aandrew: yeah that sounds good since its going to be near impossible to test all types of materials and their flags

package/com.unity.formats.usd/Documentation~/USD Import.md
lee-aandrew: TIL the colonizers use the word "anticlockwise"

package/com.unity.formats.usd/Documentation~/USD.md
vickycl: Is it clearer now?

package/com.unity.formats.usd/Documentation~/USD.md
vickycl: Yeah, I have already updated the USD.md and not the Readme.md at least once 😂 I didn't know about include, that's interesting! What a shame we can't use it 😞

package/com.unity.formats.usd/Documentation~/USD.md
lee-aandrew: yes, thank you for the change!

package/com.unity.formats.usd/Documentation~/USD Export.md
CraigHutchinson: This is an odd one... this comment was on line 48 ... confused how this happened!

> Note: Modifications to the transform of the Root GameObject are not currently reflected in the override, as Unity assumes the root in all USD files is at the origin.

package/com.unity.formats.usd/Documentation~/USD.md
vickycl: It actually isn't an option in later versions of Unity that we support, so to avoid confusion I'll remove it entirely :)

package/com.unity.formats.usd/Documentation~/USD.md
vickycl: Good point - I've added a project configuration section for this and a few other notes that aren't strict requirements but worth noting.

package/com.unity.formats.usd/Documentation~/USD Import.md
vickycl: I've added some clarification, does this help?

package/com.unity.formats.usd/Documentation~/USD Export.md
vickycl: Oh! Yes in that case I completely understand what you mean by Unity and USD being developed differently. Yes I guess it is a limitation in that case that the Exporter always sets the root to the origin. This feels a lot more specific than the limitations currently called out in that section, kind of like this is more of an implementation detail? If I was looking for the reason why my overrides were being exported at the root I'd probably look for it in this specific section, rather than a broad limitations section. What do you think?

CraigHutchinson: I definitely think you are right. I'm still learning our stack so being a bit nit-picky. I'd be fine with sorting this out in a future revision, as a 'user' isn't necessarily going to read through all the documentation. This could be addressed as FAQ and/or exporter/importer specific issues section at a later date.

package/com.unity.formats.usd/Documentation~/USD Import.md
CraigHutchinson: Definitely helps my understanding as it really helps to have some explanation of the difference.

package/com.unity.formats.usd/Documentation~/USD Export.md
```diff
@@ -0,0 +1,77 @@
...
+On Mac you will see multiple file format options - USDC and USDA.
+
+On Windows the default is .usd with a USDC format, but you can change the file ending in the Save window to specify a different file ending and encoding.
+
+
+## Export Transform Overrides
+
+Modifications to transforms in USD files can be exported as overrides to the original files. 
+
+Overrides can be exported from the USD menu with the GameObject containing the [UsdAsset](package/com.unity.formats.usd/Documentation~/USD-in-the-Editor.md#USD-Asset) component selected, which will export transform overrides for the entire hierarchy. Alternatively, you can also export just the overrides when exporting from the Recorder window by changing the 'Override Setting' to 'Export Transform Overrides Only'.
```
CraigHutchinson: Dead-link: I think this needs to use relative paths? When this is linked from Github it takes you too https://github.com/Unity-Technologies/usd-unity-sdk/blob/documentation/rewrite/package/com.unity.formats.usd/Documentation~/package/com.unity.formats.usd/Documentation~/USD-in-the-Editor.md#USD-Asset

while it should be https://github.com/Unity-Technologies/usd-unity-sdk/blob/documentation/rewrite/package/com.unity.formats.usd/Documentation~/USD-in-the-Editor.md
That's without the duplication on `package/com.unity.formats.usd/Documentation~/` 

Maybe this was required so it works on the package though and we can't have both github+package links working? I'd otherwise expected `USD-in-the-Editor.md#USD-Asset` could have worked.

vickycl: Thank you for testing this again! I think I've fixed it now as it works for me, is it fixed for you?

CraigHutchinson: Yes, I just tested it from GitHub and looks to work nicely now :)

CraigHutchinson (approved):
Looks good now

lee-aandrew (approved):
LGTM! 👍

README.md
```diff
@@ -1,242 +1,83 @@
...
+    * Variant Selection
+    * Payloads
+        * Load All and Individual
+    * Layer stacks
+* Geometry:
+    * UV Set
+    * Vertex Colour
+* Instancing:
+    * Point Instancing
+    * Scene Instancing
```
michaeljblain: Change this to Scenegraph instancing to link with USD docs.

package/com.unity.formats.usd/Documentation~/USD Import.md
```diff
@@ -0,0 +1,53 @@
...
+
+-   **Import As GameObjects**: Imports selected USD file as GameObjects directly into the Scene Hierarchy for the open scene.
+
+-   **Import as Prefab**: Imports selected USD file as a Prefab into the project folder. This method will also generate PlayableAssets and materials that can be reused. Note that the import will be much slower than for GameObjects, and the prefab must be unpacked to make any changes.
+
+-   **Import as Timeline Clip**: Imports selected USD file as a Prefab containing empty Materials and the Timeline Clip. If animation is recorded in the Timeline Clip, that can be used to animate other applicable Unity Objects.
+
+## Importing Materials
+
+To import materials from USD, firstly import the USD file using the USD menu. Then, to get the materials, change the 'Import Settings' > 'Materials' in the Inspector to 'Import Preview Surface'. Finally re-import the USD file.
```
michaeljblain: We can start this paragraph in the same way as "Importing Payloads" i.e. "Materials are not loaded by default".

package/com.unity.formats.usd/Documentation~/USD Import.md
```diff
@@ -0,0 +1,53 @@
...
+    -   SlowAndSafeAsFBX
+    -   None
+
+### Meshes
+
+Convert basis:
+- As above, meshes also go through a basis conversion as required.
+
+Triangulate meshes:
+- Converting from USD-supported mixed face meshes to a triangular mesh.
```
michaeljblain: Maybe change "mixed face" to "arbitrary polygon"

package/com.unity.formats.usd/Documentation~/USD-in-the-Editor.md
```diff
@@ -0,0 +1,94 @@
...
+* **Import Display Color (default)** - import the display color from the prim and apply this as a single color on the USD/StandardVertexColor shader.
+* **Import Preview Surface** - Import the USDPreviewSurface shader and map the textures and values to the default shader for the project’s render pipeline.
+* **None** - do not import any materials.
+
+*Payload Policy*
+Overall payload import policy;
+* **Don’t Load Payloads (default)** - Import without any payloads.
+* **Load All** - Import and load all payloads.
+ 
+ Payloads can also be individually loaded and unloaded on the individual objects.
```
michaeljblain: Change "individual" to "specific" to remove duplicate use of word.

michaeljblain: change "objects" to "prims"

michaeljblain (changes requested):
A great update to the docs. Some minor suggestions.

michaeljblain (approved):

PR #372, opened 2023-04-17
Remove guess interpolation error spam

vickycl:
## Purpose of this PR

**Ticket/Jira #:** Continuation of #352  

After landing #359, came across a case in Pixar's Kitchen demo where a particular attribute's interpolation was Vertex, but GuessInterpolation prioritizes Uniform, resulting in spamming of errors as it switches from one to the other. As both Vertex and Uniform interpolation were valid for this case, added a validity check before we decide to overwrite the input interpolation.

## Testing

Validated the error spam no longer occurs for Pixar's kitchen and that I haven't reintroduced the array overflow in ALab2.

## Overall Product Risks

**Complexity:** Minimal

**Halo Effect:** Minimal

## Additional information

**Note to reviewers:**

**Reminder:**


package/com.unity.formats.usd/Runtime/Scripts/IO/SanitizedSamples.cs
```diff
@@ -544,15 +544,24 @@ void ConvertInterpolationToFaceVarying<T>(ref T[] values, int[] vertexIndices, b
             if (values == null)
                 return;
 
-            var newInterpolation = GuessInterpolation(values.Length);
+            // We assume a certain number of values in the array for certain interpolations, so check whether this assumption will work and not cause array overflows
+            bool interpolationIsValid = false;
```
lee-aandrew: can we add a
`if interpolation == null or values == null; return false`
in the IsInterpolationValid function and then do 

```
if (!IsInterpolationValid) {
...
}
```
(line 553)

right away?

vickycl: I agree that it would be clearer if the `if interpolation == null` were inside the `IsInterpolationValid`, I've made that change :) 
However I think let's leave the values check outside that function, because 1. it adds an unnecessary function call in the case that values are null which is currently an early out (it may be compiled away, but I'm not sure), and 2. values being null has more significance than 'just' the interpolation being invalid, so logically I prefer it outside that function. What do you think of this change?

lee-aandrew: That sounds good! Thank you for the change!

lee-aandrew (approved):
LGTM!

lee-aandrew:
Tested on Unity 2019.4.26f1

Checked manually with the following assets:
- AnimalLab ALab2
- Pixar Kitchen
- J-Cube Esper Room

The interpolation warning spam is no longer present for any of these assets, though I did see one pop up during loading of ALab2 - Array overload was also not found with ALab2

Also tested NVIDIA's Attic asset, which is known to take a while to load everything for potential time out / Unity lock problems, while it does take a very long time - about 5 minutes on my laptop, it does load without any errors or warnings
I am not sure if this is a performance regression stemming from PR #359, will create a [ticket](https://jira.unity3d.com/browse/USDU-381) to check for this at a later date

PR #373, opened 2023-04-17
[USDU-201] Fix for HdrpShaderImporter.BuildMaskMap generates incorrect textures if red/green/blue/alpha texture is null 

lee-aandrew:
During import under Unity HDRP environment, BuildMaskMap function within HdrpShaderImporter can generate incorrect textures if red/green/blue/alpha texture is null

## Purpose of this PR

**Ticket/Jira #:**
[[USDU-201]](https://jira.unity3d.com/browse/USDU-201)

## Testing

N/A for now

N/A

## Overall Product Risks

**Complexity:**

**Halo Effect:**

## Additional information

**Note to reviewers:**
Please refer to the comment in our JIRA ticket

**Reminder:**


vickycl (commented):
Great, thank you! One remaining question (on [Jira](https://jira.unity3d.com/browse/USDU-201)) about whether we need to catch the red being null before the Blit. I think you've already tested it, but just making sure :)

lee-aandrew:
> Great, thank you! One remaining question (on [Jira](https://jira.unity3d.com/browse/USDU-201)) about whether we need to catch the red being null before the Blit. I think you've already tested it, but just making sure :)

I updated the JIRA ticket with the following:

```
I tried by forcing green, blue and alpha to Texture2D.grayTexture, Texture2D.whiteTexture, Texture2D.blackTexture respectively

while keeping red as null

The Blit still works and the generated MaskMap exists (it seems like in my previous tests, it was exited earlier as red green blue and alpha were all set to null)

No errors are seen
```

vickycl (approved):
Great, thanks Andrew. Would be good to add some tests in the future, but as it's HDRP I think it would need to be an extra test project, so let's not hold up on that now :)

PR #375, opened 2023-04-20
[USDU-342] Further fix for PrimvarReader varname with non connected primvar

vickycl:
## Purpose of this PR

**Ticket/Jira #:** [USDU-342](https://jira.unity3d.com/browse/USDU-342) / #353 

When PrimvarReader's varname was changed from TfToken to string by USD 21.11, the previous fix only suited variables with a connected path for it. This fix forces all imported PrimvarReaders to have a string typed varname, which still works for older versions because TfTokens can be automatically translated to strings.

This PR adds a PrimvarReaderSampleExport type that still uses a TfToken varname, so that we are still exporting shaders compatible with USD 20.08, but this should be removed in the future when we upgrade USD.

## Testing

I've modified the existing USD preview test to test both the string and TfToken types, and manually tested on a modified sample. I've tried to test:
- import and export
- connected + unconnected primvars
- "st" and non-default value "uv" for varname
- importing versions with either token or string type

**Performance Testing status:** Small fix, no anticipated performance impact.

## Overall Product Risks

**Complexity:** low

**Halo Effect:** low

## Additional information

**Note to reviewers:**

**Reminder:**

- [x ] Add entry in CHANGELOG.md _(if applicable)_


package/com.unity.formats.usd/Tests/USD.NET.Unity/UsdPreviewSurfaceTests.cs
```diff
@@ -8,76 +8,142 @@ namespace USD.NET.Unity.Tests
...
+
+            PrimvarReaderImportSample<Vector2> USDReadPrimvarReader = new PrimvarReaderImportSample<Vector2>();
+
+            ReadPrimsFromScene();
+            m_USDScene.Read(k_primvarReaderPath, USDReadPrimvarReader);
+
+            CheckShaderParams();
+            CheckMaterialValues();
+
+            Assert.AreEqual(primvarReaderTfToken.varname.defaultValue, primvarReaderTfToken.varname.defaultValue);
```
michaeljblain: Should one of these parameters refer to USDReadPrimvarReader?

package/com.unity.formats.usd/Tests/USD.NET.Unity/UsdPreviewSurfaceTests.cs
```diff
@@ -8,76 +8,142 @@ namespace USD.NET.Unity.Tests
...
+
+            PrimvarReaderSample<Vector2> USDReadPrimvarReader = new PrimvarReaderSample<Vector2>();
+
+            ReadPrimsFromScene();
+            m_USDScene.Read(k_primvarReaderPath, USDReadPrimvarReader);
+
+            CheckShaderParams();
+            CheckMaterialValues();
+
+            Assert.AreEqual(primvarReaderString.varname.defaultValue, primvarReaderString.varname.defaultValue);
```
michaeljblain: Should one of these parameters refer to USDReadPrimvarReader?

michaeljblain (changes requested):
Looks good. A couple of queries.

package/com.unity.formats.usd/Tests/USD.NET.Unity/UsdPreviewSurfaceTests.cs
vickycl: Yes :) Thanks!

package/com.unity.formats.usd/Dependencies/USD.NET.Unity/Shading/UsdPreviewSurface/PrimvarReaderSample.cs
```diff
@@ -85,4 +79,38 @@ public class Outputs : SampleBase
...
+    {
+        /// <summary>
+        /// Name of the primvar to be read from the primitive as a string.
+        /// </summary>
+        /// <remarks>
+        /// In USD 21.11 the varname was switched from a TfToken to a string. However, as a TfToken can be implicitly converted to a string,
+        /// making this a string type works for both during *import*.
+        /// </remarks>
+        [InputParameter("_Varname")]
+        public Connectable<string> varname = new Connectable<string>("st");
```
SimonBoorer: I take it st is a default value here? I'm not sure it makes much sense for it to be a default in this context?

SimonBoorer (approved):

package/com.unity.formats.usd/Dependencies/USD.NET.Unity/Shading/UsdPreviewSurface/PrimvarReaderSample.cs
vickycl: Yeah, this was so that in the case where the read had failed due to the TfToken/string mismatch there was still a value here. I'll remove it.

michaeljblain (approved):
Looks good.

PR #376, opened 2023-04-21
[draft] [USDU-370] First pass at adding relative path to UsdAsset.

vickycl:
## Purpose of this PR

**Ticket/Jira #:** [USDU-370]

When Unity project directories are moved or shared, the full path that is serialized for an imported asset causes lots of errors due to the path being different. This change attempts to, where possible, save only the relative path inside the Unity project folder, but still fallback to the full path when it's not in the project folder. We then also catch cases more cleanly when they don't exist at the stated path. This awkward code is necessary to not break existing projects when upgrading, and still support importing USD files from outside the project folder.

We could alternatively have a one time fix up style approach that happens on domain reload or something, but this seemed the simplest approach that should catch if the user fixes a missing file.

This version will be **non backwards compatible**, because the UsdAsset is 'fixed' to contain a m_relativeUsdPath, and this new field doesn't exist on older versions of the package. As many users have had too many issues with 3.0.0-exp.2 that 3.0.0-exp.3 should solve, this shouldn't be too common. However I think we should fully discuss this before landing.

## Testing

TBC

TBC

We are adding potential overhead to a getter which is certainly undesirable, however in the majority of cases it should fallback to the same cost as before- a single call to Path.GetFullPath().

## Overall Product Risks

**Complexity:** Low

**Halo Effect:** Medium

## Additional information

**Note to reviewers:**

**Reminder:**

### DIFF
--- a/package/com.unity.formats.usd/Runtime/Scripts/Behaviors/UsdAsset.cs
+++ b/package/com.unity.formats.usd/Runtime/Scripts/Behaviors/UsdAsset.cs
@@ -29,6 +29,13 @@ namespace Unity.Formats.USD
     [ExecuteInEditMode]
     public class UsdAsset : MonoBehaviour
     {
+        enum USDPathType
+        {
+            Unknown,
+            External,
+            Relative
+        }
+
         /// <summary>
         /// The length of the USD playback time in seconds.
         /// </summary>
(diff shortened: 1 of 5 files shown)


package/com.unity.formats.usd/Runtime/Scripts/Behaviors/UsdAsset.cs
```diff
@@ -209,6 +257,69 @@ private GameObject GetPrefabObject(GameObject root)
...
+        }
+
+        /// <summary>
+        /// Converts inputPath to a Relative Path, if the inputPath is inside the Project folder. If the path is not inside the project folder, returns empty string.
+        /// </summary>
+        /// <param name="inputPath">
+        /// The path to determine a path relative to the project folder from. We assume the inputPath is a full path, else this will fail.
+        /// </param>
+        private string DetermineRelativePathFromFullPath(string inputPath)
+        {
```
CraigHutchinson, line 276: May be worth looking at [System.IO.Path.GetRelativePath](https://learn.microsoft.com/en-us/dotnet/api/system.io.path.getrelativepath?view=net-7.0&viewFallbackFrom=netframework-4.0)?

vickycl: I tried that first, but it's not available on 2019.4 due to the .NET version it uses 😞

PR #380, opened 2023-05-05
Modify package CI based on templates

vickycl:
## Purpose of this PR

**Ticket/Jira #:**

Modifies the package CI to align with most recent templates and refactored version of the package.

## Testing

## Overall Product Risks

**Complexity:**

**Halo Effect:**

## Additional information

**Note to reviewers:**

**Reminder:**


.yamato/package-test.yml
```diff
@@ -0,0 +1,136 @@
...
+  name : Test {{ editor.version }} on {{ platform.name }}
+  agent:
+    type: {{ platform.type }}
+    image: {{ platform.image }}
+    flavor: {{ platform.flavor}}
+  variables:
+    UPMCI_PKG: "upm-ci-utils@stable"
+  commands:
+      - npm install {% if platform.name == "win" %}"%UPMCI_PKG%"{% else %}"$UPMCI_PKG"{% endif %} -g --registry {{ registry.npm }}
+      - upm-ci package test -u {{ editor.version }} --package-path package/com.unity.formats.usd --extra-create-project-arg=-upmNoDefaultPackages --extra-utr-arg="--api-profile=NET_4_6"
```
SimonBoorer: Should this have --type package-tests?

.yamato/package-test.yml
```diff
@@ -0,0 +1,136 @@
...
+  name : Test Recorder {{ editor.version }} on {{ platform.name }}
+  agent:
+    type: {{ platform.type }}
+    image: {{ platform.image }}
+    flavor: {{ platform.flavor}}
+  variables:
+    UPMCI_PKG: "upm-ci-utils@stable"
+  commands:
+      - npm install {% if platform.name == "win" %}"%UPMCI_PKG%"{% else %}"$UPMCI_PKG"{% endif %} -g --registry {{ registry.npm }}
+      - upm-ci project test  --project-path TestProject/Usd-Recorder --type project-tests --unity-version {{ editor.version }} --extra-utr-arg="--api-profile=NET_4_6"
```
SimonBoorer: Should this have --extra-create-project-arg=-upmNoDefaultPackages?

SimonBoorer: It might also be good to reorder the arguments and use -u {{ editor.version }} to keep this consistent with the package tests

.yamato/package-test.yml
```diff
@@ -0,0 +1,136 @@
...
+  agent:
+    type: {{ platform.type }}
+    image: {{ platform.image }}
+    flavor: {{ platform.flavor}}
+  variables:
+    UPMCI_PKG: "upm-ci-utils@stable"
+  commands:
+    - npm install {% if platform.name == "win" %}"%UPMCI_PKG%"{% else %}"$UPMCI_PKG"{% endif %} -g --registry {{ registry.npm }}
+    - upm-ci package test --unity-version {{ backend.editor }} --package-path package/com.unity.formats.usd
+      --platform standalone --backend {{ backend.name }} --extra-utr-arg="--testfilter=StandaloneSanityTest {{ backend.apiprofile }}"
```
SimonBoorer: Same comments as above

.yamato/package-test.yml
```diff
@@ -0,0 +1,136 @@
...
+    - .yamato/package-test.yml#test_{{platform.name}}_{{editor.version}}
+    - .yamato/package-test.yml#test_recorder_{{platform.name}}_{{editor.version}}
+{% endfor %}
+{% endfor %}
+{% for backend in test_backends %}
+{% for platform in test_platforms %}
+    - .yamato/package-test.yml#test_standalone_{{ platform.name }}_{{ backend.editor }}_{{ backend.name }}
+{% endfor %}
+{% endfor %}
+    - .yamato/formatting.yml#formatting
```
SimonBoorer: Perhaps this would be good to rename package-formatting.yml for consistency? There is a package-coverage.yml in the templates, we could try and contribute this back?

.yamato/package.metafile
```diff
@@ -0,0 +1,58 @@
...
+  - version: 2019.4
+
+promotion_test_editors:
+  - version: trunk
+  - version: 2019.4
+
+test_platforms:
+  - name: win
+    type: Unity::VM
+    image: package-ci/win10:v4
```
SimonBoorer, line 16: This is a change from the previous package-ci/win10:stable and package-ci/macos-12:stable images which is a good thing as stable shouldn't be used anymore. I see the templates are using MacOS 13 rather than 12, do you think we should upgrade?

SimonBoorer (approved):

.yamato/package-test.yml
vickycl: Good point, it looks like the [default is isolation tests](https://internaldocs.unity.com/package_development/upm-ci/commands/package-context/#test).
> Executes the complete suite of tests required for publishing (package validation suite, clean console, package has tests, and package tests)

These extra tests that will no longer be executed are part of vetting-tests, which runs during 'validation'. I think that's a better place for them to run, so I'll switch this to just package tests :)

.yamato/package-test.yml
vickycl: Hmm, I'm not really sure if it needs the --extra-create-project-arg=-upmNoDefaultPackages 🤔 It's using the Test Project Usd-Recorder, so I wouldn't expect it to be creating a new Unity project to test in, but the documentation still calls it out 😅 I'll add it in and if the tests still all pass we can go with it

.yamato/package-test.yml
vickycl: Contributing this back is a great idea. I'll find out who I need to speak to to suggest it :)

.yamato/package.metafile
vickycl: Hmm, good question. I think in terms of ensuring compatibility with minimum requirements for 2019.4 we should be aiming for as low a MacOS version as possible as 2019.4 supports 10.12 and up. However I think MacOS 12 is the earliest we can actually support because of certain graphics test issues, in which case maybe using the older doesn't really matter as it's not the min spec anyway...

michaeljblain (approved):
Looks good. Nice to see the yamato templates update as well.

PR #385, opened 2023-05-19
Add Editor Analytics for package.

vickycl:
## Purpose of this PR

**Ticket/Jira #:** [USDU-384](https://jira.unity3d.com/browse/USDU-384)

Links: [Spike document](https://docs.google.com/document/d/1hl0kTdK-9Gqkkbm1ywfAHAHT7z_OcYvBmPgvUayDW4A/edit#) | [Editor Data Collection form](https://docs.google.com/spreadsheets/d/1bOhBgplhthCxSfIIJI2dTxp6uGW1SOYVNbqk6S_55mw/edit#gid=1274740786)

Begins adding basic analytics to understand package usage. Includes:
- Initialization of the plugin
- Import events + relevant context:
    - success
    - file format
    - time taken
    - features used
- Export events + relevant context:
    - success
    - file format
    - is it only overrides
    - time taken
- Recorder export

This PR doesn't capture _everything_ we would like to, as called out in the [doc](https://docs.google.com/document/d/1hl0kTdK-9Gqkkbm1ywfAHAHT7z_OcYvBmPgvUayDW4A/edit#heading=h.a5hv0kmbhb0). It's mostly an MVP to build upon. Payloads and Variants events standout as features that might be relatively simple but useful to collect.

## Testing

Currently attempting to write automated tests. Manually tested that the basic workflows collate the data correctly using the Analytics Debugger, and viewed the data on BigQuery, all looks ok.

None yet.

## Overall Product Risks

**Complexity:** medium

**Halo Effect:** low

## Additional information

**Note to reviewers:**

**Reminder:**


package/com.unity.formats.usd/Editor/Scripts/UsdBuildPostProcess.cs
```diff
@@ -45,20 +46,28 @@ public static void OnPostprocessBuild(BuildTarget target, string pathToBuiltProj
                 return;
             }
 
-            if (!Directory.Exists(destination))
+            try
             {
-                Directory.CreateDirectory(destination);
+                if (!Directory.Exists(destination))
```
michaeljblain: Can tidy this up a bit to only have 1 call to Directory.CreateDirectory.

package/com.unity.formats.usd/Runtime/Scripts/IO/Geometry/InstanceImporter.cs
```diff
@@ -163,6 +163,8 @@ public static void BuildSceneInstances(PrimMap primMap, SceneImportOptions optio
 
                 primMap.AddInstance(goInstance);
             }
+
+            primMap.ContainsPointInstances = instanceCount > 0;
```
michaeljblain: Could we move this up a level and just set it once if we have >0 point instancer prims (rather than calling it for every point instancer). My assumption was that we'd just capture whether point instancer prims exist within the imported usd  and not check the point instancer data itself.

michaeljblain (commented):
Looks like a solid first pass. Some minor comments. Just wondering how the known issues can be addressed.

package/com.unity.formats.usd/Runtime/Scripts/IO/Geometry/InstanceImporter.cs
vickycl: Done :)

package/com.unity.formats.usd/Editor/Scripts/UsdBuildPostProcess.cs
vickycl: Oh yeah, whoops!

package/com.unity.formats.usd/Common/UsdEditorAnalytics.cs
```diff
@@ -0,0 +1,199 @@
+# if ENABLE_CLOUD_SERVICES_ANALYTICS && UNITY_EDITOR
```
SimonBoorer: Copyright notice should probably be at the beginning of the file

package/com.unity.formats.usd/Common/UsdEditorAnalytics.cs
```diff
@@ -0,0 +1,199 @@
...
+        {
+# if USE_EDITOR_ANALYTICS
+            // If EditorAnalytics are disabled, don't waste cycles doing any of the set up
+            if (!EditorAnalytics.enabled)
+                return false;
+
+            if (sUsdEditorAnalyticsEvents[eventName])
+                return true;
+
+            AnalyticsResult result = EditorAnalytics.RegisterEventWithLimit(eventName, k_MaxEventsPerHour, k_MaxNumberOfElements, k_VendorKey);
```
SimonBoorer, line 142: What are the other possible results for AnalyticsResult? Should we log if registration fails?

package/com.unity.formats.usd/Common/UsdEditorAnalytics.cs
```diff
@@ -0,0 +1,199 @@
...
+                TimeTakenS = timeTakenS
+            };
+
+            EditorAnalytics.SendEventWithLimit(k_UsageEventName, data);
+# endif
+        }
+
+        public static void SendImportEvent(string fileExtension, float timeTakenS, ImportResult result)
+        {
+            Debug.Log($"Send event called");
```
SimonBoorer: Remove me?

package/com.unity.formats.usd/Runtime/InitUsd.cs
```diff
@@ -56,9 +60,13 @@ public static bool Initialize()
             catch (System.Exception ex)
             {
                 Debug.LogException(ex);
+                analyticsTimer.Stop();
+                UsdEditorAnalytics.SendUsageEvent(false, analyticsTimer.ElapsedMilliseconds * 0.001f);
```
CraigHutchinson: Minor: Could use API `analyticsTimer.Elapsed.TotalSeconds` here and elsewhere

package/com.unity.formats.usd/Common/UsdEditorAnalytics.cs
```diff
@@ -0,0 +1,199 @@
...
+
+        // USD Recorder Export Analytics
+        const string k_RecorderExportEventName = "uUSDFileRecorderExport";
+
+        struct RecorderExportAnalyticsData
+        {
+            public string FileExtension;
+            public bool ExportSucceeded;
+            public bool OnlyOverrides;
+            public int FrameCount;
```
SimonBoorer, line 118: Should we be recording time taken here too or is that not valid in this context?

package/com.unity.formats.usd/Editor/Scripts/Behaviors/UsdMenu.cs
```diff
@@ -14,9 +14,9 @@
 
 using System.IO;
 using System.Linq;
-using UnityEngine;
 using UnityEditor;
 using USD.NET;
+using Debug = UnityEngine.Debug;
```
SimonBoorer: Not used?

package/com.unity.formats.usd/Runtime/InitUsd.cs
```diff
@@ -12,10 +12,11 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
+using System.Diagnostics;
 using System.Runtime.CompilerServices;
-using UnityEngine;
 using USD.NET;
 using USD.NET.Unity;
+using Debug = UnityEngine.Debug;
```
SimonBoorer, line 18: Not used?

package/com.unity.formats.usd/Runtime/Scripts/IO/Scene/ImportHelpers.cs
```diff
@@ -8,7 +9,9 @@
 using UnityEngine;
 using USD.NET;
 using USD.NET.Unity;
+using Debug = UnityEngine.Debug;
```
SimonBoorer: Not used?

package/com.unity.formats.usd/Runtime/Scripts/IO/Scene/SceneImporter.cs
```diff
@@ -14,11 +14,13 @@
 
 using System.Collections;
 using System.Collections.Generic;
+using System.Diagnostics;
 using UnityEngine;
 using UnityEngine.Profiling;
 using USD.NET;
 using USD.NET.Unity;
 using Unity.Jobs;
+using Debug = UnityEngine.Debug;
```
SimonBoorer: Not used?

package/com.unity.formats.usd/Runtime/Scripts/IO/Scene/PrimMap.cs
```diff
@@ -48,19 +48,23 @@ public struct InstanceRoot
...
+        // Flags for import analytics
+        public bool ContainsPointInstances { get; set; }
+
         // Normal objects in the hierarchy.
         private Dictionary<SdfPath, GameObject> m_prims = new Dictionary<SdfPath, GameObject>();
 
         // Objects at the root of an instanced sub-tree.
         // Instances may be found in masters as well.
-        private Dictionary<SdfPath, InstanceRoot> m_instanceRoots = new Dictionary<SdfPath, InstanceRoot>();
+        public Dictionary<SdfPath, InstanceRoot> InstanceRoots = new Dictionary<SdfPath, InstanceRoot>();
```
SimonBoorer: Seems like this change is no longer necessary with the addition of ContainsPointInstances?

package/com.unity.formats.usd/Runtime/Scripts/Behaviors/UsdAsset.cs
```diff
@@ -590,14 +590,22 @@ public void ExportOverrides(Scene sceneInWhichToStoreTransforms)
...
+
             if (overs == null)
             {
+                UsdEditorAnalytics.SendExportEvent(Path.GetExtension(sceneToReference.usdFullPath), .0f, false, onlyOverrides: true);
                 return;
             }
 
+            bool success = false;
+
+            System.Diagnostics.Stopwatch analyticsTimer = new System.Diagnostics.Stopwatch();
```
SimonBoorer: Would be good to standardise on either using System.Diagnostics or System.Diagnostics.Stopwatch

same with Unity.Formats.USD.UsdEditorAnalytics.ImportResult

SimonBoorer (commented):
General approach looks good, some minor nitpicks

package/com.unity.formats.usd/Dependencies/USD.NET/collections/SampleCollection.cs
```diff
@@ -29,6 +29,8 @@ public class SampleCollection<T> : IEnumerable<SampleEnumerator<T>.SampleHolder>
         private SdfPath[] m_paths;
         private Scene m_scene;
 
+        public int GetLength() => m_paths.Length;
```
CraigHutchinson: Minor: Length is a property and could be defined as such: `public int Length { get =>  m_paths.Length; }`

package/com.unity.formats.usd/Runtime/Scripts/IO/Scene/PrimMap.cs
CraigHutchinson: Comment: Not sure if this is intended to be public, was private before change and coudln't spot usage outside of this.

SimonBoorer:
Another thought - I think you can use timer.Elapsed.Seconds rather than timer.ElapsedMilliseconds * 0.001f - but might it make more sense to capture time in milliseconds anyway?

CraigHutchinson (approved):
No major observations, only  minor changes that may be addressed

package/com.unity.formats.usd/Dependencies/USD.NET/collections/SampleCollection.cs
vickycl: My reasoning for calling it 'GetLength' was so that it was explicit that it was retrieving something from a member, and not to be confused with an Array.Length. I might be overthinking this though. Do you think making it a property would be nicer?

package/com.unity.formats.usd/Editor/Scripts/Behaviors/UsdMenu.cs
vickycl: It's used in a function that I haven't changed, where previous it just used the UnityEngine include instead. I'll revert this change :)

package/com.unity.formats.usd/Runtime/InitUsd.cs
vickycl: EDIT: this is because adding System.Diagnostics added ambiguity for Debug, so in this case just saves changing all the existing Debug.Log() calls to UnityEngine.Debug.Log().

package/com.unity.formats.usd/Runtime/InitUsd.cs
vickycl: Ah, I missed that API :)

package/com.unity.formats.usd/Runtime/Scripts/Behaviors/UsdAsset.cs
vickycl: What do you mean by this? Do you mean specifically in the 'using's sometimes I've done it and sometimes I haven't?

package/com.unity.formats.usd/Common/UsdEditorAnalytics.cs
vickycl: Hmm, I guess it would be valid, while exporting via the recorder we do some work every frame so we could keep a cumulative total of the time taken to export to USD over all of the frames.

package/com.unity.formats.usd/Common/UsdEditorAnalytics.cs
vickycl: The other results are the 'reasons' it can fail, eg UnsupportedPlatform. We could debug log it for the user, but I'm not sure whether they could necessarily do anything about it, and as the analytics has failed to initialize then we can't report it back to ourselves. 

Maybe it's worth Debug.Logging anyway, so that we could get slightly more informative bug reports if a user decided to report it.

vickycl: I've added logging for users on both the registering and sending of events :)

package/com.unity.formats.usd/Runtime/Scripts/IO/Scene/ImportHelpers.cs
vickycl: In this case I've swapped the `using System.Diagnostics;` for `using Stopwatch = System.Diagnostics.Stopwatch;` for consistency and removed the `using Debug = UnityEngine.Debug;` - is that better?

package/com.unity.formats.usd/Runtime/Scripts/IO/Scene/PrimMap.cs
vickycl: Yep- forgot to revert it. Thanks :)

vickycl: You're right- I forgot to revert this when I switched to the ContainsPointInstances flag instead :)

package/com.unity.formats.usd/Runtime/Scripts/Behaviors/UsdAsset.cs
SimonBoorer: Yeah pretty much, change looks good thanks!

SimonBoorer (approved):

package/com.unity.formats.usd/Dependencies/USD.NET/collections/SampleCollection.cs
CraigHutchinson: Its more just a C# convention as the 'get' (and 'set') prefix is handled as part of the C# language using properties. 

Really boils down to your personal preference and the .Net library itself isn't consistent but once you set the name its harder to change. Some points for official reference
https://learn.microsoft.com/en-us/previous-versions/dotnet/netframework-1.1/bzwdh01d(v=vs.71)?redirectedfrom=MSDN#properties-vs-methods

vickycl: Yep, seems like in this case Length is actually the way to go. Thanks :)

SimonBoorer (approved):

package/com.unity.formats.usd/Common/UsdEditorAnalytics.cs
```diff
@@ -22,6 +22,14 @@
 
 namespace Unity.Formats.USD
 {
+    public enum ImportType
+    {
+        Initial,
+        Reimport,
```
michaeljblain: To avoid confusion with `bool IsReImport` in the Analytics class perhaps we call this something like "Replace" or "ForceRebuild" or similar.

michaeljblain (commented):
Looks good. Given the work is already done we could actually store output the import type as part of the analytics (instead of a bool).

package/com.unity.formats.usd/Runtime/InitUsd.cs
```diff
@@ -61,12 +61,12 @@ public static bool Initialize()
             {
                 Debug.LogException(ex);
                 analyticsTimer.Stop();
-                UsdEditorAnalytics.SendUsageEvent(false, analyticsTimer.ElapsedMilliseconds);
+                UsdEditorAnalytics.SendUsageEvent(false, (float)analyticsTimer.Elapsed.TotalMilliseconds);
```
michaeljblain: Is there a reason why we cast this to float instead of just sending through the original double value?

michaeljblain (approved):
Looks good to me. Just one question about float vs double.

michaeljblain (changes requested):
This seems to now correctly indicate import failure for our test case. Longer term we should address error handling in a more thorough way. 
Some minor feedback: 
* the commit message doesn't indicate the purpose of the inclusion of the `SceneTemplateSettings.json` file.
* I think we still should store the ms as a double instead of casting to float.

vickycl:
@michaeljblain 

> This seems to now correctly indicate import failure for our test case. Longer term we should address error handling in a more thorough way. Some minor feedback:
> 
> * the commit message doesn't indicate the purpose of the inclusion of the `SceneTemplateSettings.json` file.

This was an accident :) thanks for pointing it out!

> * I think we still should store the ms as a double instead of casting to float.

Done

michaeljblain (approved):
Thanks for the changes.

package/com.unity.formats.usd/Documentation~/USD.md
```diff
@@ -114,3 +114,8 @@ We do not currently support the following:
     * Blend Shapes
 * General:
     * Custom prims
+
+
+## Analytics
+
+We collect high level analytics of the features used in the package, to help us understand adoption and prioritize future development work. You can see exactly what data points are collected in `Common\UsdEditorAnalytics.cs`.
```
CraigHutchinson, line 121: Comment: Maybe something for future, but maybe there is a contact email for analytics at Unity so as/if corporate entities have concerns over our collections they can make queries to address the concern that could arise from analytics being present. i.e. Opt-in or out GDPR sort of stuff

package/com.unity.formats.usd/Runtime/Scripts/IO/Scene/ExportHelpers.cs
```diff
@@ -65,7 +66,7 @@ public static Scene InitForSave(string filePath)
             scene.Save();
 
             analyticsTimer.Stop();
-            UsdEditorAnalytics.SendExportEvent(Path.GetExtension(scene.FilePath), analyticsTimer.ElapsedMilliseconds * 0.001f, success);
+            UsdEditorAnalytics.SendExportEvent(Path.GetExtension(scene.FilePath), (float)analyticsTimer.Elapsed.TotalMilliseconds, success);
```
CraigHutchinson: Moderate: Unnecessary float cast

package/com.unity.formats.usd/Runtime/Scripts/IO/Scene/SceneImporter.cs
```diff
@@ -278,14 +277,19 @@ static SceneImporter()
             SceneImportOptions importOptions)
         {
             UsdEditorAnalytics.ImportResult importResult = UsdEditorAnalytics.ImportResult.Default;
+            importResult.ImportType = importOptions.ImportType;
             if (scene == null)
             {
-                UsdEditorAnalytics.SendImportEvent("", .0f, importResult);
+                UsdEditorAnalytics.SendImportEvent("", 0.0f, importResult);
```
CraigHutchinson: Trivial: Could use 0 or 0.0 as in other places as this is now 'double' type

CraigHutchinson (commented):
Looks good but a really minor (float) cast left over from earlier revision

SimonBoorer (approved):

package/com.unity.formats.usd/Runtime/Scripts/IO/Scene/ExportHelpers.cs
vickycl: Good spot :)

package/com.unity.formats.usd/Documentation~/USD.md
vickycl: These are just standard analytics the same as any other in the Unity Editor, as described in the official documentation: https://docs.unity3d.com/Manual/EditorAnalytics.html

PR #386, opened 2023-05-19
[USDU-312] Test Code refactor

lee-aandrew:
## Purpose of this PR

The test code is messy in terms of:
- un-uniform methods of calling test usd assets, 
- file structure of scripts and test assets
- various placement of helper functions that should be under a single TestUtility type of script
- various test asset GUIDs that are in each test script, would be better kept in a single location so duplicate variable init can be reduced
- hard coded strings that can be fixed
- existence of un-used function definitions
- Some tests are in runtime tests when they should be in editmode tests

**Ticket/Jira #:**
USDU-312

- InitUsd script is no longer under the package's Runtime folder, it has been moved over to a new "Common" folder
-- To match this, a new Common assembly definition has been created
- All helper functions within BaseFixture.cs has been moved into TestUtility function, as not all children of BaseFixture were using these functions
- all USD files used for testing is now referenced by their GUIDs
-- All test USD files' GUIDs are now under a single script, "TestAssetData"
-- All USD (related) test asset files are now moved into /Test/Common/Assets
- Removed a few hardcoded strings that is better kept in the TestAssetData script or the TestUtility script
- Removed un-used test helper functions
- Moved some ImportHelperTest (Runtime) test cases to ImportHelperTests (EditorMode)

## Testing

N/A: was a refactoring task

## Overall Product Risks

**Complexity:**

Min

**Halo Effect:**

## Additional information

**Note to reviewers:**

**Reminder:**


package/com.unity.formats.usd/Tests/Common/Unity.Formats.USD.Tests.Common.asmdef
```diff
@@ -19,9 +20,7 @@
         "nunit.framework.dll"
     ],
     "autoReferenced": true,
-    "defineConstraints": [
-        "UNITY_INCLUDE_TESTS"
```
vickycl: I think we need to keep this so that the APIs in the assembly aren't accessible externally.

package/com.unity.formats.usd/Tests/Common/TestAssetData.cs
```diff
@@ -1,26 +1,116 @@
...
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
 namespace Unity.Formats.USD.Tests
 {
     public static class TestAssetData
```
vickycl: This feels like slightly excessive encapsulation here. Having to write TestAssetData.GUID.PrimType.collectionsUsd feels a bit inaccessible to read and non-obvious when using intellisense to find it when writing new tests. Maybe we could reduce the naming to be two layers max? Eg in this case TestData_USDFileGUIDs.CollectionsUsd. 

I'm not sure there's much value extracting const strings that are only used in one place out of their respective files, such as ShaderParams, as this seems to be adding unnecessary complexity.

package/com.unity.formats.usd/Common/Unity.Formats.USD.Common.asmdef
vickycl, line 1: I've added this in my PR too, so will need to be careful when we merge them :)

package/com.unity.formats.usd/Tests/Common/Assets/Material/Textured/Textures/textured_metallic.png.meta
vickycl, line 1: Makes perfect sense to move all of these files out of a file structure naming that assumed a specific test, however I don't think it should be called 'Assets' as I think that's a reserved name for the Assets folder and there's a risk of issues arising from that- I think keeping it as 'Data' is safer. 'Materials/Textured/Textures' seems excessive- maybe just 'Materials/Textures' :)

package/com.unity.formats.usd/Tests/Editor/ImportHelpersTests.cs
```diff
@@ -1,52 +1,139 @@
...
+            Assert.IsFalse(usdObject.activeInHierarchy);
+            Assert.IsTrue(usdObject.activeSelf, "The USD Scene is self-inactive when imported under an inactive parent");
         }
 
         [Test]
         public void ImportAsPrefabTest_SceneClosedAfterImport()
         {
-            ImportHelpers.ImportAsPrefab(m_scene, GetPrefabPath());
-            Assert.IsNull(m_scene.Stage, "Scene was not closed after import.");
+            var scene = TestUtility.CreateTestUsdScene(ArtifactsDirectoryFullPath);
```
vickycl, line 103: Could this be a one time setup instead of every test?

package/com.unity.formats.usd/Tests/Editor/Unity.Formats.USD.Tests.Editor.asmdef
```diff
@@ -5,10 +5,10 @@
         "USD.NET.Unity",
         "Unity.Formats.USD.Runtime",
         "Unity.Formats.USD.Editor",
-        "Unity.Formats.USD.Tests.Common"
-    ],
-    "optionalUnityReferences": [
-        "TestAssemblies"
```
vickycl, line 11: What's the reasoning for changing this?

package/com.unity.formats.usd/Tests/Runtime/SanitizedSamplesTests.cs
```diff
@@ -235,13 +233,12 @@ public void UniformToFaceVarying_ValuesAreCorrect()
...
     {
         GameObject m_usdRoot;
-        string m_usdGUID = "c5046ed8700c010469c4c557353bc241"; // GUI of mesh_instances.usda
 
         [SetUp]
         public void SetUp()
         {
             InitUsd.Initialize();
-            var usdPath = Path.GetFullPath(AssetDatabase.GUIDToAssetPath(m_usdGUID));
+            var usdPath = Path.GetFullPath(AssetDatabase.GUIDToAssetPath(TestAssetData.GUID.Instancer.instancedUsd));
```
vickycl: I think this should have 'mesh' in the name, as 'instanced' is too vague. Alternatively keep the definition in this file as it's not used anywhere else so doesn't need to be abstracted :)

vickycl (commented):
Really nice clean up! 

There's a couple of aspects that I think might be over-abstracted which will make maintaining and writing new tests more difficult, so watch out for that, eg really long folder hierarchies, multiple levels of encapsulation for const strings, strings abstracted from the only file they're used in.

package/com.unity.formats.usd/Tests/Common/Assets/Material/Textured/Textures/textured_metallic.png.meta
lee-aandrew: - I renamed "Assets" to "Data" as per ur suggestion + to better match the folder naming used in "Runtime" and "Editor" folders
- I moved up the contents of "Textured" folder to its parent "Materials" (so its "Materials/Textures")

package/com.unity.formats.usd/Tests/Common/TestAssetData.cs
lee-aandrew: I moved out the GUIDS into its own script TestDataGuids.cs
this reduces the lvl of encapsulation by 1, I dont want to reduce it to be less than that, as its purpose was to mimic the physical location of the file it is referencing

ex. TestDataGuids.Material.SimpleMaterialUsd == Tests/Common/Data/Material/simpleMaterialTest.usd

I deleted the original TestAssetData as it was only left with const strings

package/com.unity.formats.usd/Tests/Editor/ImportHelpersTests.cs
lee-aandrew: We could, but it probably isnt the best option here since there are some tests that close the scene, while in most cases we need it to be opened

We could also maybe move it to a SetUp funciton, but not all tests will use it, so its not the best solution either i think

I think it would be best to leave it as is

package/com.unity.formats.usd/Tests/Editor/Unity.Formats.USD.Tests.Editor.asmdef
lee-aandrew: Good point, im not sure why it was deleted, I added it back

package/com.unity.formats.usd/Tests/Runtime/SanitizedSamplesTests.cs
lee-aandrew: Changed variable name to MeshInstancesUsd

package/com.unity.formats.usd/Tests/USD.NET/OverrideTests.cs
```diff
@@ -1,4 +1,4 @@
-// Copyright 2018 Jeremy Cowles. All rights reserved.
+// Copyright 2023 Unity Technologies. All rights reserved.
```
michaeljblain, line 2: I think this should be added "in addition" to the original copyright notice rather than instead. See section 4.d of the Apache 2.0 license:

> You may add Your own copyright statement to Your modifications and may provide additional or different license terms and conditions for use, reproduction, or distribution of Your modifications, or for any such Derivative Works as a whole, provided Your use, reproduction, and distribution of the Work otherwise complies with the conditions stated in this License.

michaeljblain (changes requested):

SimonBoorer:
Is there any particular benefit to referencing test files by GUID instead of relative paths or is this just the Unity way of doing things?

lee-aandrew:
> Is there any particular benefit to referencing test files by GUID instead of relative paths or is this just the Unity way of doing things?

There is the benefit of not having to worry about test file names being changed or its directory being changed

SimonBoorer (approved):

package/com.unity.formats.usd/Tests/USD.NET/OverrideTests.cs
lee-aandrew: Fixed!

package/com.unity.formats.usd/Tests/Editor/UsdPrefabTests.cs
```diff
@@ -33,7 +31,7 @@ public class Reload : UsdPrefabTests
             [Ignore("USDU-276")]
             public void UsdPrefab_ForceReload_Succeed(MaterialImportMode mode)
             {
-                ImportAsPrefab(k_USDZGUID);
+                ImportAsPrefab(TestDataGuids.PrimType.ComponentPayloadUsd);
```
vickycl: This should be 'TestDataGuids.PrimType.ComponentPayloadUsd**z**' so that it's super clear on file type

vickycl (commented):
Great work- one really small fix for future clarity. It might be relevant feedback to other places too- where the test file is USDZ I think this should be specified in the string for both easier reuse and tracking down future failures.

package/com.unity.formats.usd/Tests/Editor/UsdPrefabTests.cs
lee-aandrew: Fixed!

vickycl (approved):
👍

package/com.unity.formats.usd/Tests/Common/TestUtility.cs
```diff
@@ -0,0 +1,155 @@
...
+    public static class TestUtility
+    {
+#if UNITY_EDITOR
+        public static USDScene OpenUSDSceneWithGUID(string usdGUID, UsdStage.InitialLoadSet loadSet = UsdStage.InitialLoadSet.LoadNone)
+        {
+            var usdPath = Path.GetFullPath(AssetDatabase.GUIDToAssetPath(usdGUID));
+            return OpenUSDSceneWithFullPath(usdPath, loadSet);
+        }
+
+        // TODO: If materialImportMode = MaterialImportMode.ImportPreviewSurface, it creates all the texture2d files on the root assets
```
michaeljblain: Maybe we can move this TODO to a ticket.

package/com.unity.formats.usd/Tests/USD.NET/OverrideTests.cs
```diff
@@ -1,4 +1,4 @@
-// Copyright 2018 Jeremy Cowles. All rights reserved.
+// Copyright 2018 Jeremy Cowles & 2023 Unity Technologies. All rights reserved.
```
michaeljblain: WE should split this into 2 lines as before.

michaeljblain (approved):
Some minor suggestions. Otherwise looks good.

package/com.unity.formats.usd/Tests/Common/TestUtility.cs
lee-aandrew: added JIRA ticket ID to the TODO

package/com.unity.formats.usd/Tests/USD.NET/OverrideTests.cs
lee-aandrew: split them up as separate lines for each of the copyright

lee-aandrew:
Thanks for the reviews @SimonBoorer @vickycl @michaeljblain !

PR #391, opened 2023-06-15
Adds Define Constraints for the editor assembly files

lee-aandrew:
## Purpose of this PR
Adds Define Constraints for the editor assembly files
The Editor assembly files that were added last push were missing the define constraints required

**Ticket/Jira #:**

## Testing

## Overall Product Risks

**Complexity:**

**Halo Effect:**

## Additional information

**Note to reviewers:**
Now that i think about it, i dont know why our samples' assembly definitions have the "UNITY_INCLUDE_TESTS" Define constraints when they arent tests - we should perhaps delete the constraint or have a different constraint such as "USD_SAMPLES"

**Reminder:**

### DIFF
--- a/package/com.unity.formats.usd/Samples/ExportMesh/Editor/Unity.Formats.USD.ExportMesh.Editor.asmdef
+++ b/package/com.unity.formats.usd/Samples/ExportMesh/Editor/Unity.Formats.USD.ExportMesh.Editor.asmdef
@@ -1,5 +1,6 @@
 {
     "name": "Unity.Formats.USD.ExportMesh.Editor",
+    "rootNamespace": "",
     "references": [
         "Unity.Formats.USD.ExportMesh"
     ],
@@ -11,7 +12,9 @@
     "overrideReferences": false,
     "precompiledReferences": [],
     "autoReferenced": true,
-    "defineConstraints": [],
+    "defineConstraints": [
+        "UNITY_INCLUDE_TESTS"
+    ],
     "versionDefines": [],
     "noEngineReferences": false
 }
--- a/package/com.unity.formats.usd/Samples/ImportMesh/Editor/Unity.Formats.USD.ImportMeshEditor.asmdef
+++ b/package/com.unity.formats.usd/Samples/ImportMesh/Editor/Unity.Formats.USD.ImportMeshEditor.asmdef
@@ -1,5 +1,6 @@
 {
     "name": "Unity.Formats.USD.ImportMeshEditor",
+    "rootNamespace": "",
     "references": [
         "Unity.Formats.USD.ImportMesh",
         "Unity.Formats.USD.Runtime",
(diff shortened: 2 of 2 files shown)


package/com.unity.formats.usd/Samples/ExportMesh/Editor/Unity.Formats.USD.ExportMesh.Editor.asmdef
```diff
@@ -11,7 +12,9 @@
     "overrideReferences": false,
     "precompiledReferences": [],
     "autoReferenced": true,
-    "defineConstraints": [],
+    "defineConstraints": [
+        "UNITY_INCLUDE_TESTS"
```
CraigHutchinson, line 16: I'm not upto date on this one, so this is making the samples depend on the define for Tests being present while this is under Samples. I noticed you have mentioned this in the notes for reviewers also so I will confirm and also noted this https://forum.unity.com/threads/info-on-unity_include_tests-define.890095/#post-5933573

michaeljblain: I'd like to find out more about why this is the right way to fix this issue. Is it now just excluding the sample?

vickycl: I _think_ this is making sure that the Samples assemblies aren't being included in a Player build. It seems to be unrelated to test code. It's needed here because it's set in the `Unity.Formats.USD.ExportMesh` assembly that this one depends on, and without it this assembly is being compiled into the player unnecessarily and thus having compile errors about missing the other assembly. Whether this is the 'right way', I'm not sure, it's just the way that previous devs have implemented it. I will have a play around and see if I can see what the reasoning is for it in the higher level asmdef, as I don't see it in any of the small sample of other package samples I've looked at.

CraigHutchinson: Concur with @vickycl. One of the types that are issue is"\Samples\ImportMesh\ImportMeshExample.cs" which isn't part of the editor build, seems odd or maybe this is an issue arising from something 'legacy'

vickycl: I'm not really following what you mean there Craig- let's discuss in the Foundation sync later :) I have a feeling the correct fix would be to make all of the Samples asmdefs Editor-only, but I can't reproduce the error reliably so I haven't been able to test this and that it doesn't break anything else. I don't know if any of the test code reuses sample code either, which would not be possible in that case.

CraigHutchinson: Sounds like a plan. I'm no longer clear on what I stated as I see that 'UNITY_INCLUDE_TESTS' should effectively always be defined for Editor or Test-Assemblies. However it appears this may not be the case the Package code so could depend how this is included into a project (TBC)

vickycl (approved):
Seems reasonable as a temporary fix, we should do a proper review of these settings for next release :)

PR #396, opened 2023-07-03
[USDU-438] Fix export from project folder

vickycl:
## Purpose of this PR

**Ticket/Jira #:** #395 / [USDU-438](https://jira.unity3d.com/browse/USDU-438)

The fix for not allowing multiple objects with the same name introduced a check for any siblings of an object having the same name. In the case of a root GameObject, it uses the scene the GO is in to find any siblings. If it is not in a scene (as is the case for a Prefab selected in the project window), this scene is null and the export fails with an error.

This PR adds a simple check for whether the Scene is valid. I believe it is fine to early out in this case, because having two files with the same name selected in the project view doesn't seem to be possible.

There is a weird edge case where perhaps a user exports two prefabs with the same name through script. I haven't figured out a fix for that case yet.

## Testing

Added a new automated test case, manually verified fix. Testing flagged a separate issue with exporting skeletons from a prefab, so I raised a [ticket](https://jira.unity3d.com/browse/USDU-439) for that.

Shouldn't have any impact on performance.

## Overall Product Risks

**Complexity:** 1

**Halo Effect:** 1

## Additional information

**Note to reviewers:**

**Reminder:**

### DIFF
--- a/package/com.unity.formats.usd/CHANGELOG.md
+++ b/package/com.unity.formats.usd/CHANGELOG.md
@@ -1,5 +1,13 @@
 # Changes in usd-unity-sdk for Unity
 
+## [3.0.0-exp.5] - Unreleased
+### Features
+
+### Bug Fixes
+- Fixed export for Prefabs from the Project Window.
+
+### Changed
+
 ## [3.0.0-exp.4] - 2023-06-21
 ### Features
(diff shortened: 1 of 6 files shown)


SimonBoorer (approved):
I verified that ExportPrefabFromProjectWindow_ExportsWithoutErrors fails without the change, LGTM!

PR #398, opened 2023-07-10
[USDU-407] Add automated tests for Analytics 

lee-aandrew:
## Purpose of this PR
Analytics have been added for our USD package giving us various information about the usage of the package
We thought it would have be needed to have some automated test cases for the analytics

**Ticket/Jira #:**
[USDU-407]

## Testing

Adds functional testcases for the analytics for the following parts:
- Import
- Export
- Re-import (refresh / update)

Could not add planned Recorder analytics automated testing as a lot of recorder aspect is private / internal, and unless I use reflections I cannot instantiate a lot of the classes that I would need to set up the test
LMK if you guys think its needed

## Overall Product Risks

**Complexity:**

**Halo Effect:**

## Additional information

**Note to reviewers:**

**Reminder:**


SimonBoorer:
Are you aware the tests are failing on 2023.1 and newer?
> Library\PackageCache\com.unity.formats.usd.tests@090deec1ff8c\Tests\Editor\EditorAnalyticsBaseFixture.cs(8,7): error CS0246: The type or namespace name 'Newtonsoft' could not be found (are you missing a using directive or an assembly reference?)

SimonBoorer (approved):

package/com.unity.formats.usd/Tests/Common/BaseFixture.cs
```diff
@@ -75,6 +75,8 @@ public string GetPrefabPath(string prefabName = null, bool resource = false)
                 prefabName += ".prefab";
             }
 
+            var test = Path.Combine(ArtifactsDirectoryRelativePath, resource ? "Resources" : "", prefabName);
```
michaeljblain: Are these 2 lines just for your personal debugging and can they be removed?

package/com.unity.formats.usd/Tests/Common/BaseFixture.cs
```diff
@@ -126,7 +137,14 @@ public void CleanupTestArtifacts()
...
             if (Directory.Exists(ArtifactsDirectoryFullPath))
             {
-                Directory.Delete(ArtifactsDirectoryFullPath, true);
+                try
+                {
+                    Directory.Delete(ArtifactsDirectoryFullPath, true);
+                }
+                catch
+                {
+                    Debug.Log("Artifact Clean up has failed - This should not happen in most cases, but even if so, the test case should not be affected.");
```
michaeljblain, line 59: Can we feed back further detail on the failure by also logging the exception?

package/com.unity.formats.usd/Tests/Editor/EditorAnalyticsBaseFixture.cs
```diff
@@ -0,0 +1,149 @@
+# if UNITY_2023_2_OR_NEWER
```
michaeljblain: Are we restricted to >= 2023.2 due to Vicky's Analytics fix on trunk? Or is there some other reason? Should we consider adding a note here about why it is 2023.2 or newer?

michaeljblain (changes requested):
Some suggestions. Looks like tests are still failing on trunk.

package/com.unity.formats.usd/Tests/Editor/EditorAnalyticsBaseFixture.cs
lee-aandrew: yes the restriction is due to her fix only being on 2023.2+

package/com.unity.formats.usd/Tests/Common/BaseFixture.cs
lee-aandrew: yes i forgot i left them in, sorry! Deleted!

TestProject/Usd-Development/Packages/manifest.json
```diff
@@ -1,11 +1,12 @@
 {
   "dependencies": {
     "com.unity.analytics": "3.5.3",
+    "com.unity.editor-analytics-debugger": "0.1.0",
```
vickycl: Is this ok in 2019.4?

lee-aandrew:
~Stuck due to problem on the analytics side - Mike Stowell is looking into the issue~

Requires Mike's fixes: [Link](https://github.cds.internal.unity3d.com/unity/unity/pull/33470)

.yamato/package-test.yml
```diff
@@ -13,7 +13,11 @@ test_{{ platform.name }}_{{ editor.version }}:
     UPMCI_PKG: "upm-ci-utils@stable"
   commands:
       - npm install {% if platform.name == "win" %}"%UPMCI_PKG%"{% else %}"$UPMCI_PKG"{% endif %} -g --registry {{ registry.npm }}
-      - upm-ci package test -u {{ editor.version }} --package-path package/com.unity.formats.usd --type package-tests --extra-utr-arg="--api-profile=NET_4_6"
+      {% if editor.version == "trunk" %}
+        - upm-ci package test -u {{ editor.version }} --package-path package/com.unity.formats.usd --type package-tests --extra-utr-arg="--api-profile=NET_4_6" --extra-editor-arg=-editorAnalyticsTestMode
```
lee-aandrew: This is behind an if statement as there was a bug in the analytics side where if this was set for versions less than trunk, it would crash the editor

vickycl: I don't think we should specify 'trunk' here as it isn't robust for the future. Eventually trunk will be newer that 2023 and the '23 LTS won't be tested/ the tests will run without this editor arg being set. I'm not sure what the correct answer is though :)

vickycl: does the `--extra-editor-arg=-editorAnalyticsTestMode` alleviate the need for UNITY_THISISABUILDMACHINE? ie does it set the runmode to test?

package/com.unity.formats.usd/Tests/Common/Data/Invalid/Kiki_Blendshape.usd
vickycl, line 1: I can't see what is included here, but it's quite a big asset to include for the sake of a test, and I'm also not sure whether it can be shared publicly in this way?

package/com.unity.formats.usd/Tests/Editor/EditorAnalyticsBaseFixture.cs
```diff
@@ -0,0 +1,150 @@
...
+        }
+
+        [SetUp]
+        public void ResetEventList()
+        {
+            DebuggerEventListHandler.ClearEventList();
+            InitUsd.Initialize();
+        }
+
+        public IEnumerator WaitForUsdAnalytics<T>(string expectedType, System.Action<AnalyticsEvent> expectedEvent, float attemptFrameCountLimit = 300) where T: UsdAnalyticsEvent
```
vickycl: This frame count limit is very high- are you sure it needs to be this high?

package/com.unity.formats.usd/Tests/Editor/EditorAnalyticsBaseFixture.cs
```diff
@@ -0,0 +1,150 @@
...
+            {
+                AnalyticsEvent expectedAnalyticsEvent = new AnalyticsEvent();
+                foreach (var concattedEvent in DebuggerEventListHandler.fetchEventList())
+                {
+                    if (!concattedEvent.Contains(expectedType))
+                    {
+                        continue;
+                    }
+
+                    Debug.Log($"Raw Log: {concattedEvent}");
```
vickycl: Please remove the logging in every test run- it adds a lot of noise and some overhead. If the logging is necessary, maybe just do it on test failure?

package/com.unity.formats.usd/Tests/Editor/EditorImportAnalyticsTests.cs
```diff
@@ -0,0 +1,212 @@
...
+            ImportAs(m_scene, importMethod);
+
+            AnalyticsEvent expectedEvent = null;
+            yield return (WaitForUsdAnalytics<UsdAnalyticsEventImport>(UsdAnalyticsTypes.Import, (AnalyticsEvent waitedEvent) => {
+                expectedEvent = waitedEvent;
+            }));
+
+            Assert.IsNotNull(expectedEvent);
+
+            var expectedUsdEventMsg = ((UsdAnalyticsEventImport)expectedEvent.usd).msg;
```
vickycl: Nitpick: This isn't the 'expected', it's the 'actual', right?

package/com.unity.formats.usd/Tests/Editor/EditorReImportAnalyticsTests.cs
```diff
@@ -0,0 +1,149 @@
...
+            yield return (WaitForUsdAnalytics<UsdAnalyticsEventReImport>(UsdAnalyticsTypes.ReImport, (AnalyticsEvent waitedEvent) => {
+                reImportEvent = waitedEvent;
+            }));
+
+            Assert.IsNotNull(reImportEvent);
+            var reImportUsdEventMsg = ((UsdAnalyticsEventReImport)reImportEvent.usd).msg;
+
+            Assert.IsTrue(reImportUsdEventMsg.Succeeded, "Expected True for re-import Successful");
+            Assert.Greater(reImportUsdEventMsg.TimeTakenMs, 0, "Expected Time Taken MS should be greater than 0 for Re-Import");
+            Assert.AreEqual(importUsdEventMsg.IncludesSkel, reImportUsdEventMsg.IncludesSkel);
```
vickycl, line 143: I think it would be better to check the Reimport result against a constant expected value instead of relying on the Import message, because if something went wrong that broke both Import and Reimport, comparing the two against each other wouldn't show anything wrong

package/com.unity.formats.usd/Common/UsdEditorAnalytics.cs
```diff
@@ -40,7 +42,11 @@ public static class UsdEditorAnalytics
         // Universal USD Analytics
         const string k_UsageEventName = "USDPackageUsage";
 
-        struct UsageAnalyticsData
+        public interface IUsdAnalyticsData
```
vickycl: This change seems to be unnecessary now since you implemented the UsdAnalyticsEvent class in the tests?

package/com.unity.formats.usd/Tests/Editor/EditorAnalyticsBaseFixture.cs
```diff
@@ -0,0 +1,150 @@
...
+        }
+    }
+
+    public class AnalyticsEvent
+    {
+        public SharedAnalyticsEvent shared;
+        public UsdAnalyticsEvent usd;
+    }
+
+    public class SharedAnalyticsEvent
```
vickycl: The shared event is no longer used anywhere- can it be removed now?

package/com.unity.formats.usd/Tests/Editor/EditorExportAnalyticsTests.cs
```diff
@@ -0,0 +1,133 @@
...
+
+            AnalyticsEvent exportEvent = null;
+            yield return (WaitForUsdAnalytics<UsdAnalyticsEventExport>(UsdAnalyticsTypes.Export, (AnalyticsEvent waitedEvent) => {
+                exportEvent = waitedEvent;
+            }));
+
+            Assert.IsNotNull(exportEvent);
+            var exportUsdEventMsg = ((UsdAnalyticsEventExport)exportEvent.usd).msg;
+
+            Assert.IsTrue(exportUsdEventMsg.Succeeded, "Expected True for re-import Successful");
```
vickycl: think this should be 'export successful'

package/com.unity.formats.usd/Tests/Editor/EditorExportAnalyticsTests.cs
```diff
@@ -0,0 +1,133 @@
...
```
vickycl: As above :)

package/com.unity.formats.usd/Tests/Editor/EditorExportAnalyticsTests.cs
```diff
@@ -0,0 +1,133 @@
...
+
+            AnalyticsEvent exportEvent = null;
+            yield return (WaitForUsdAnalytics<UsdAnalyticsEventExport>(UsdAnalyticsTypes.Export, (AnalyticsEvent waitedEvent) => {
+                exportEvent = waitedEvent;
+            }));
+
+            Assert.IsNotNull(exportEvent);
+            var exportUsdEventMsg = ((UsdAnalyticsEventExport)exportEvent.usd).msg;
+
+            Assert.IsFalse(exportUsdEventMsg.Succeeded, "Expected Fail for re-import Successful");
```
vickycl: As above :)

package/com.unity.formats.usd/Tests/Editor/EditorAnalyticsBaseFixture.cs
vickycl: In which case can we remove a layer of abstraction and use the UsdAnalyticsEvent directly instead of the AnalyticsEvent wrapper? To simplify lines like `var exportUsdEventMsg = ((UsdAnalyticsEventExport)exportEvent.usd).msg;` a bit, as they're hard to parse

vickycl: for eg `var exportUsdEventMsg = ((UsdAnalyticsEventExport)exportEvent).eventData;` in that case would be easier to reason at a glance what it's getting

package/com.unity.formats.usd/Tests/Editor/EditorAnalyticsBaseFixture.cs
```diff
@@ -0,0 +1,150 @@
...
+    {
+        // check if the class for this is in Unity source (if its a bit convoluted just make it)
+        object common;
+    }
+
+    public class UsdAnalyticsEvent
+    {
+        public string type;
```
vickycl, line 114: Am I understanding correctly that this actually stores the event name string eg `editor.USDFileImport`, or is it actually the C# type name of the event? If it is the former I'd suggest a more explicit variable name :)

vickycl (commented):
Small fixes and clean up suggested, but otherwise good so far!

TestProject/Usd-Development/Packages/manifest.json
vickycl: Oops, this is leftover from when I initially started the review 😂

package/com.unity.formats.usd/Tests/Editor/EditorAnalyticsBaseFixture.cs
```diff
@@ -0,0 +1,150 @@
...
+
+    public class AnalyticsEvent
+    {
+        public SharedAnalyticsEvent shared;
+        public UsdAnalyticsEvent usd;
+    }
+
+    public class SharedAnalyticsEvent
+    {
+        // check if the class for this is in Unity source (if its a bit convoluted just make it)
```
vickycl: What does this comment mean?

package/com.unity.formats.usd/Tests/Editor/EditorAnalyticsBaseFixture.cs
lee-aandrew: nono its the former, its the event name, i'll change it to something like "name" or "eventName"

.yamato/package-test.yml
lee-aandrew: I realized i set the "UNITY_THISISABUILDMACHINE" but it never got passed over to the yamato call, and i believe our test machines already have the "UNITY_THISISABUILDMACHINE" set to true

package/com.unity.formats.usd/Tests/Common/Data/Invalid/Kiki_Blendshape.usd
lee-aandrew: Thats a good point, i just needed an invalid USD file for a failure, in this case blendshapes, i'll look for another asset

package/com.unity.formats.usd/Tests/Editor/EditorAnalyticsBaseFixture.cs
lee-aandrew: It can be removed, but i meant to check if theres a class with the same structure as "SharedAnalyticsEvent" already defined within Unity, and if so, use that instead of creating it here

package/com.unity.formats.usd/Tests/Editor/EditorImportAnalyticsTests.cs
lee-aandrew: good point - will change the names

.yamato/package-test.yml
lee-aandrew: Yeah I think that will have to be changed in the future, once neutron stuff becomes the new trunk (iirc 2023 is the last "classic" unity before neutron)

If there is a 2023.4 of "classic" Unity that becomes the new trunk, in that case we will have to add 2023.3 as part of this if statement

package/com.unity.formats.usd/Tests/Editor/EditorAnalyticsBaseFixture.cs
lee-aandrew: actually i cant change this as the result json's variable is called "type" and as far as i know Unity's JsonUtility does not have something similar to `JsonProperty`

package/com.unity.formats.usd/Common/UsdEditorAnalytics.cs
lee-aandrew: Good point, will revert

package/com.unity.formats.usd/Tests/Editor/EditorAnalyticsBaseFixture.cs
lee-aandrew: Changed it to use c#'s stopwatch  for shorter + more accurate measurement

package/com.unity.formats.usd/Tests/Editor/EditorReImportAnalyticsTests.cs
lee-aandrew: The point of this specific test case is that for the initial import and re-import the properties are equal
If theres a problem with the initial import, it would have been caught with the import tests, and if theres a problem with the re-import it would have been caught in the re-import specific tests
This testcase is to make sure that on re-import, some properties do not change for simple files like the test file

PR #406, opened 2023-08-04
[USDU-476] Improve error messaging for Apple Silicon users.

vickycl:
## Purpose of this PR

**Ticket/Jira #:** USDU-476

This package doesn't currently support Apple Silicon native editors, but this isn't always clear to users and they may run into unclear exceptions. This PR improves exception handling in this case.

I don't particularly like this solution, I think it would be far better to detect Apple Silicon earlier and grey out all USD functionality including the menus, the components, etc. However that is a much larger piece of work for something that should be temporary.

## Testing

Tested on an M1 Mac in Rosetta emulation mode and Silicon native to check that the plugin still works on the former and useful errors are included on the latter.

N/A

## Overall Product Risks

**Complexity:** Low

**Halo Effect:** Low

## Additional information

**Note to reviewers:**

**Reminder:**

### DIFF
--- a/package/com.unity.formats.usd/Documentation~/index.md
+++ b/package/com.unity.formats.usd/Documentation~/index.md
@@ -10,8 +10,8 @@ This package includes a set of libraries designed to support the use of USD in C
 
 ## Requirements
 
-* Windows / OSX (Intel only)
-    * Apple Silicon architectures are not currently supported by this package. To use this package on M1 devices, you must use an x64 install of the Unity Editor.
+* Windows / OSX (*Intel only*)
+    * Apple Silicon architectures are not currently supported by this package. To use this package on M1 and M2 devices, you must use an x64 install of the Unity Editor.
 * Unity version: 2019.4 and up
 
 ## Project Configuration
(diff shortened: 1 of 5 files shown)


package/com.unity.formats.usd/Documentation~/index.md
```diff
@@ -40,7 +40,10 @@ Once the USD package is installed, a USD menu will appear, enabling you to easil
 
 ## Package Limitations
 
-* The USD plugin bundles are not currently code signed, so will need to be manually signed or removed from standalone player builds.
+* The USD plugin bundles are not currently code signed, so will need to be manually signed. 
+    * In some cases on MacOS devices, this may mean manually removing the 'com.apple.quarantine' meta flag from the UsdCs.bundle plugin, eg: run `xattr -d com.apple.quarantine <UsdCs.bundle path>`.
```
michaeljblain: Should we indicate to users where in the package folder subtree the bundle lives?

michaeljblain (approved):
One minor suggestion. Otherwise looks good and behaves as expected when tested.

package/com.unity.formats.usd/Documentation~/index.md
vickycl: Done :)

PR #408, opened 2023-08-07
[USDU-472] Fix USD component removal on Prefabs

vickycl:
## Purpose of this PR

**Ticket/Jira #:** #405 / [USDU-472](https://jira.unity3d.com/browse/USDU-472)

There were bugs in the 'detach USD' and 'remove USD' workflows for Prefabs that this PR fixes:
1. When editing the Prefab Asset from the Project Window, both of these workflows resulted in errors because we were not forcing the 'DestroyImmediate' to destroy assets.
2. When editing the Prefab Asset in prefab editing mode, these workflows did not set the scene as dirty, so the changes were not saved and therefore not persisting. This change registers the action as undo-able, with the side effect of tracking changes to trigger the scene-save.

## Testing

Tested locally that all four expected situations act as expected:
1. Import as Game Object -> remove components should work
2. Import as Prefab -> open in Inspector from Project Window -> remove components should work, instances of the prefab should have same change
3. Import as Prefab -> open in Prefab Editing mode -> remove components should work, instances of the prefab should have same change
4. Import as Prefab -> add instance to scene and modify this instance -> should not be able to make destructive changes.

@nicolaspopravka confirmed on #405 that this fixes the issue for them :) 

There may be a slight increase in time taken for the removal process, but before it wasn't doing the expected work so this is unavoidable.

## Overall Product Risks

**Complexity:** low

**Halo Effect:** low

## Additional information

**Note to reviewers:**

**Reminder:**

### DIFF
--- a/package/com.unity.formats.usd/CHANGELOG.md
+++ b/package/com.unity.formats.usd/CHANGELOG.md
@@ -7,6 +7,7 @@
 ### Bug Fixes
 - Fixed export for Prefabs from the Project Window.
 - Fixed export of URP base and occlusion maps for Lit shader.
+- Fixed USD component removal workflows for Prefabs.
 
 ### Changed
 - Renamed USD.md in the Documentation folder to index.md to match documentation generation standards for landing pages.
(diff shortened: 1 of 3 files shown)


CraigHutchinson (approved):
Code changes look good

SimonBoorer (approved):

